PR #97, opened 2020-10-08
Add macOS and Windows to build matrix

jayqi:


jayqi:
@pjbull it kind of seems to me from the Windows build failure logs that boto3 is not successfully authenticating with AWS? Any ideas?

pjbull:
Maybe an env var thing? You might do a check that (1) they get set correctly on windows from the work flow, and (2) if they are set correctly, the python process can find them with `os.getenv`.

jayqi:
@pjbull so what I'm observing is that something is weird where boto3 isn't authenticating properly in our tests on Windows. The environment variables exist and are not empty. The part that fails isn't cloudpathlib code, it's the part that uses boto3 directly to upload the test assets to the bucket to set up each test. It's failing to authenticate both with the default session that picks up the env vars, and when I explicitly create a session with the env vars read in with `os.getenv`. I haven't been able to find any useful information about what's going on. 

It works fine on Ubuntu and macOS. The Azure tests have no problem. I made [this additional script](https://github.com/drivendataorg/cloudpathlib/pull/97/files#diff-81b9c30a2f2970b4d0b3ded5641dee90aa564333d210e72d7b3ae8e568decc6f) that tries running boto3 reads/writes and `S3Path` read/writes and they run without any problem. I'm totally stumped.

In the interest of moving forward, and getting the conda-forge package published, I'm thinking maybe we just disable the tests on Windows for now and open an issue.

jayqi:
I'm closing this to keep the history but making the changes that don't involve testing in #98

PR #100, opened 2020-10-18
Fix Windows test suite

### LINKED ISSUE #99: Fix tests on Windows
From #97, we encountered a few bugs in the tests for Windows. #98 has the mocked and live test steps disabled for the Windows CI. 

1. There are some parts of the tests where local paths are used to write `CloudPath` paths, that need to go through a `PurePosixPath` conversion. See the diff in #97 for those places. 

2. For the live tests, S3 rig setup was failing to upload test assets to the test bucket. It appears to be some kind of authentication problem. Environment variables appear to be generally working, and read/write in a separate script outside the tests had no problems. See some of the changes in #98 to see what was tried. 

```
E           boto3.exceptions.S3UploadFailedError: Failed to upload D:\a\cloudpathlib\cloudpathlib\tests\assets\dir_0\file0_0.txt to ***/TQxxmXB3Vr5dQCLcSgaHxB-test_s3_specific-test_s3path_properties/dir_0/file0_0.txt: An error occurred (SignatureDoesNotMatch) when calling the PutObject operation: The request signature we calculated does not match the signature you provided. Check your key and signing method.
```

(from [this build](https://github.com/drivendataorg/cloudpathlib/runs/1247915278?check_suite_focus=true))

3. There are potentially some issues with the mocked backends. It's possible they overlap with above issues (see 1 above). See [this build's logs](https://github.com/drivendataorg/cloudpathlib/runs/1228827652?check_suite_focus=true)

pjbull:
live tests and mocked tests passing on windows dev machine

Closes #99

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,6 +2,8 @@
 
 black
 flake8
+ipython
+jupyter
 matplotlib
 mkdocs>=1.1
 mkdocs-jupyter
@@ -14,6 +16,7 @@ pytest
 pytest-cases
 pytest-cov
 python-dotenv
+pywin32; sys_platform == 'win32'
 shortuuid
 tabulate
 wheel
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,6 @@ jobs:
       - name: Run mocked tests
         run: |
           make test
-        if: matrix.os != 'windows-latest'
 
       - name: Run live tests
         run: |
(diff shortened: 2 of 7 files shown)


requirements-dev.txt
```diff
@@ -14,6 +16,7 @@ pytest
 pytest-cases
 pytest-cov
 python-dotenv
+pywin32; sys_platform == 'win32'
```
jayqi, line 19: Huh, was the VM you were using win32? Didn't even know 32-bit stuff was still around.

jayqi (approved):

requirements-dev.txt
pjbull: 😆 I don't think that it was, but there's a bunch for historical reasons I think the the os level api on windows is often referred to as "win32"

Issue #43, opened 2020-08-27
Make mypy pass


pjbull:
Current state:

```
$ mypy cloudpathlib
cloudpathlib/cloudpath.py:13: error: "Type[PurePosixPath]" has no attribute "resolve"
cloudpathlib/backends/s3/s3backend.py:193: error: "Type[S3Backend]" has no attribute "S3Path"
cloudpathlib/backends/azure/azblobbackend.py:200: error: "Type[AzureBlobBackend]" has no attribute "AzureBlobPath"
Found 3 errors in 3 files (checked 11 source files)
```

jayqi:
Closed by #102

PR #102, opened 2020-11-06
Simplify resolve

### LINKED ISSUE #45: Don't monkeypatch Path.resolve (and simplify resolve)
Currently we have this [heavily-modified copy of `resolve` from pathlib](https://github.com/drivendataorg/cloudpathlib/blob/604fdec5b82a3c2665d977827d49265529a7ae55/cloudpathlib/_vendored.py#L16) which gets used when [`CloudPath` dispatches to `PurePosixPath` methods](https://github.com/drivendataorg/cloudpathlib/blob/604fdec5b82a3c2665d977827d49265529a7ae55/cloudpathlib/cloudpath.py#L398-L405). 

This has some potential to be simplified. We should revisit this issue when there is a more robust test suite. 

### (A) Don't monkeypatch, just call as a function

Current this method is monkeypatched onto `PurePosixPath` but this has the side-effect that it actually modifies `PurePosixPath` for users. This isn't really necessarily, because no internal methods of `PurePosixPath` will call it, since it doesn't normally exist. We can just call it as a function in `CloudPath` methods as needed.

### (B) Simplify logic

Resolve does a few things:

1. Make path absolute by ensuring a single leading `/`
2. Collapse multiple slashes into one, e.g., `a//b` -> `a/b`
3. Handle `.` and `..`

(2) already happens for free on `PurePosixPath` instantiation so it's not really necessary:

```
>>> PurePosixPath("foo//////bar")
PurePosixPath('foo/bar')
```

It's not clear that (3) is a desirable mechanism for `CloudPaths` and it may be a bug that we do something with them.

If all we need is to handle (1), then this can be done more simply.

jayqi:
- Simplifies `resolve`: remove commented stuff and unneeded function layer
- Don't monkeypatch resolve onto `PurePosixPath`: we don't want to change that class for users
- Get mypy passing and add it to lint

Resolves #45, #43


pjbull (approved):
Change looks good to me. Might be nice to have some more tests for resolve (could even be standalone unit tests).

jayqi:
I added a couple.

PR #107, opened 2020-11-12
LocalPath implementations and documentation for test mocking

### LINKED ISSUE #106: Mocked CloudPath class(es)
It could be useful to have a CloudPath class that uses the local filesystem. People who use cloudpathlib can then monkeypatch it in for a real CloudPath class in their tests.

jayqi:
- New abstract subclasses `LocalPath` and `LocalClient` that use the local filesystem as "cloud" storage. 
- New classes `LocalAzureBlobClient`, `LocalAzureBlobPath`, `LocalS3Client`, `LocalS3Path` that are intended to be monkeypatch substitutes for testing (UPDATE: +`LocalGSClient`, `LocalGSPath`)
- New documentation article ~["Testing with cloudpathlib"](https://deploy-preview-107--gallant-agnesi-5f7bb3.netlify.app/testing_with_cloudpathlib/)~ ["Testing code that uses cloudpathlib"](https://deploy-preview-107--gallant-agnesi-5f7bb3.netlify.app/testing_mocked_cloudpathlib/) with examples of how to patch things for tests
- New test rigs that run the full test suite on the new local classes

Resolves #106


jayqi:
@pjbull I think this is ready for review!

cloudpathlib/cloudpath.py
```diff
@@ -658,7 +657,7 @@ def _upload_local_to_cloud(self, force_overwrite_to_cloud: bool = False):
         # if cloud does not exist or local is newer or we are overwriting, do the upload
         if (
             not self.exists()  # cloud does not exist
-            or (self._local.stat().st_mtime > self.stat().st_mtime)
+            or (self._local.stat().st_mtime >= self.stat().st_mtime)
```
jayqi, line 660: @pjbull I was getting a test failure because two writes were happening so fast that they produced the same timestamp. Let me know what you think about this. Alternatively, we could add a sleep to the test, but I think it's a genuine situation that could come up in the wild too.

jayqi: Nevermind, I think this was fixed by #112. This is changed back to the way it was before.

jayqi:
@pjbull I rebased this PR and added support for the new GS classes.

pjbull (approved):
🎉 

Would be good to add a bullet to "why cloudpathlib" at the top of the readme that says something like "Testability: We provide `Local*` classes that can be used to easily mock cloud storage in your unit tests."

Issue #96, opened 2020-10-08
Hello from fsspec!


martindurant:
I see that we have some goals in common. [fsspec](https://filesystem-spec.readthedocs.io/en/latest/) is a mature library in use by Dask, pandas, zarr and others. However, we *don't* have a good implementation of `pathlib` on our API. Would you like to join forces to implement pathlib-on-fsspec, using the code you have already developed?

xref: https://github.com/intake/filesystem_spec/issues/434

another attempt to do this, explicitly for fsspec: https://github.com/datarevenue-berlin/drfs

pjbull:
👋  @martindurant sounds cool! 

There are probably a few level of integration that could work well. We intend to make it pretty easy to add custom clients/backends by implementing a relatively small surface area. We're adding more docs on this, but here's the `boto3` one as an example. Implementing each of the functions there should make it simple to enable all of the `pathlib` API that we support:

https://github.com/drivendataorg/cloudpathlib/blob/8692c596d98292c4ca44fd2e654effb595db1427/cloudpathlib/s3/s3client.py

Option (1):
I could see a `FsspecClient` as part of `cloudpathlib` that if you `pip install cloudpathlib[fsspec]` then it uses the fsspec client for any of the schemes that fsspec supports. There's probably some details in the coupling between the specific kind of path object we return (e.g., `S3Path`) and the `Client` object that would need to be worked out, but that all seems pretty doable.

Option (2):
You could that same extension work for a custom `Client` and `CloudPath` object within `fsspec` and the expose the factory methods as you see fit in your API.

Other thoughts on interoperability? There's definitely a big chunk of cool things in `fsspec` that we don't want to duplicate (chunks/buffering, compressing cache, async, etc.).

martindurant:
I would say those two options sound pretty similar, except for where the code lives - which I am not too worried about at all. You are probably more likely to get to it, if you have time to experiment - but seeing that this was already requested at fsspec, maybe someone else in the community contributes some pathlib interface anyway.

Do you think there's an argument for S3FileSystem, GCSFileSystem and FTPFileSystem from fsspec-land to supplant the implementations you have or are developing? They may be more complete and reduce the amount of work you need to do.

jayqi:
@martindurant @pjbull 

I took a quick stab at making an `FsspecClient` and `FsspecPath` classes that is backed by an fsspec `FileSystem` in #109. 

```
from cloudpathlib.fsspec import FsspecClient, FsspecPath
[...]

PR #109, opened 2020-11-16
[WIP] Proof of concept fsspec integration

jayqi:
Implementation of `FsspecClient` and `FsspecPath` that work with an fsspec concrete filesystem implementation. 

Follow up on #96 

---

- New abstract `FsspecClient`, `FsspecPath` classes. Each registered implementation in [fsspec.registry](https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/registry.html#ReadOnlyRegistry) has subclasses autogenerated, e.g., `S3Path`, `GCSPath`, `HTTPPath`. 
- `FsspecPath` will automatically dispatch to correct subclass, the same way `CloudPath` does. This dispatching is isolated to `FsspecPath`; `CloudPath` still only dispatches to original implementations. 
  - Running environment variable `FSSPEC_MODE=1` will _replace_ `CloudPath`'s dispatching so that it will dispatch to fsspec subclasses instead of the normal ones. 
- New extras. 
  - `[fsspec]` will install `fsspec`. 
  - Each extras from `fsspec` has been copied to `[fsspec-<name>]`, e.g.`fsspec-s3fs`. 
  - `[fsspec-all]` will install _all_ `fsspec` extras.
  - original `[all]` does not include any fsspec extras

### DIFF
--- a/cloudpathlib/__init__.py
+++ b/cloudpathlib/__init__.py
@@ -1,3 +1,4 @@
+import os
 import sys
 
 from .azure.azblobclient import AzureBlobClient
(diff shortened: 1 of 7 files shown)


pjbull:
See discussion in #96

PR #111, opened 2020-12-22
Reduce network overhead of _refresh_cache

### LINKED ISSUE #110: Slow performance for small files
Currently, `_refresh_cache` is called as part of a lot of methods to ensure the local cache is up-to-date. 

Unfortunately, `_refresh_cache` makes a whole bunch of network requests to fetch metadata from the cloud storage service. In fact, it looks like a run through `_refresh_cache` that **_doesn't_** download a new copy still needs to make 4 of these requests, while a run through that _does_ download makes a total of 6 requests. On my current internet connection, one of these requests is 350-400 ms. This means cloud paths may take a minimum of 1-2 sec for doing any method that hits `_refresh_cache` for a file that exists in cloud storage, no matter how small the file is. 

Screen Shot 2020-11-19 at 5 04 49 PM

https://github.com/drivendataorg/cloudpathlib/blob/8b230c3c7dd339c439a9b6f5af666c745ce15724/cloudpathlib/cloudpath.py#L612-L650

pjbull:
We used to do a lot of network calls in _refresh_cache, which slowed down scripts that looked at lots of files. This change makes it so `_refresh_cache` only needs one network call.

 - [x] Add a `NoStat` exception if we can't get stats for a particular path (two cases: could be directory or could not exist)
 - [x] Get stats once at top of call

Downsides:
 - We used to be able to provide a specific error if somehow you tried to cache a directory instead of a file. I don't think this will happen, so it probably is not worth the additional network call.

On a slow connection, it seems ~4x faster after removing the additional calls. (First `_refresh_cache` is slow because it actually downloads the file on my slow connection)

BEFORE:
Screen Shot 2020-12-21 at 11 30 52 AM

AFTER:
Screen Shot 2020-12-21 at 11 25 52 AM

Closes #110

### DIFF
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -89,21 +89,31 @@ class S3Client(Client):
             return "file" if cloud_path._local.is_file() else "dir"
 
         # get first item by listing at least one key
-        try:
-            res = next(_ for _ in self._s3_file_query(cloud_path))
-        except StopIteration:
+        s3_obj = self._s3_file_query(cloud_path)
+
+        if s3_obj is None:
             return None
 
         # since S3 only returns files when filtering objects:
         # if the first item key is equal to the path key, this is a file; else it is a dir
-        return "file" if res.key == cloud_path.key else "dir"
+        return "file" if s3_obj.key == cloud_path.key else "dir"
 
     def _exists(self, cloud_path: S3Path) -> bool:
-        return len([_ for _ in self._s3_file_query(cloud_path)]) > 0
+        return self._s3_file_query(cloud_path) is not None
 
     def _s3_file_query(self, cloud_path: S3Path):
         """Boto3 query used for quick checks of existence and if path is file/dir"""
-        return self.s3.Bucket(cloud_path.bucket).objects.filter(Prefix=cloud_path.key).limit(1)
+        return next(
+            (
+                obj
+                for obj in (
+                    self.s3.Bucket(cloud_path.bucket)
+                    .objects.filter(Prefix=cloud_path.key)
+                    .limit(1)
+                )
+            ),
+            None,
+        )
 
     def _list_dir(self, cloud_path: S3Path, recursive=False) -> Iterable[S3Path]:
         bucket = self.s3.Bucket(cloud_path.bucket)
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -84,10 +84,6 @@ class S3Client(Client):
         if not cloud_path.key:
             return "dir"
 
-        # short circuit files/dirs that are already in the cache
-        if cloud_path._local.exists():
-            return "file" if cloud_path._local.is_file() else "dir"
-
         # get first item by listing at least one key
         s3_obj = self._s3_file_query(cloud_path)
 
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -84,26 +84,26 @@ class S3Client(Client):
         if not cloud_path.key:
             return "dir"
 
+        # short circuit files/dirs that are already in the cache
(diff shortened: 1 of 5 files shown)


jayqi (approved):
This looks like a reasonable solution to me.

pjbull:
Two updates, and I think this is ready to go if all the tests pass.

(1) In testing, found that `_upload_local_to_cloud` also had repeated stat calls, so refactored that.

(2) While looking for speed improvements, found that filtering the objects list made both our `_exists` and `_is_file_or_dir` calls on the S3 client faster. Here are some test runs:

image

cloudpathlib/s3/s3client.py
```diff
@@ -84,26 +84,26 @@ def _is_file_or_dir(self, cloud_path: S3Path) -> Optional[str]:
...
-            obj = self.s3.ObjectSummary(cloud_path.bucket, cloud_path.key)
-            obj.get()
-            return "file"
-        except self.client.exceptions.NoSuchKey:
-            prefix = cloud_path.key
-            if prefix and not prefix.endswith("/"):
-                prefix += "/"
+        # short circuit files/dirs that are already in the cache
+        if cloud_path._local.exists():
+            return "file" if cloud_path._local.is_file() else "dir"
```
jayqi, line 89: What happens if the cloud copy changed and the local cache is out of date?

cloudpathlib/s3/s3client.py
```diff
@@ -84,26 +84,26 @@ def _is_file_or_dir(self, cloud_path: S3Path) -> Optional[str]:
...
+        # if the first item key is equal to the path key, this is a file; else it is a dir
+        return "file" if res.key == cloud_path.key else "dir"
 
     def _exists(self, cloud_path: S3Path) -> bool:
-        return self._is_file_or_dir(cloud_path) in ["file", "dir"]
+        return len([_ for _ in self._s3_file_query(cloud_path)]) > 0
+
+    def _s3_file_query(self, cloud_path: S3Path):
+        """Boto3 query used for quick checks of existence and if path is file/dir"""
+        return self.s3.Bucket(cloud_path.bucket).objects.filter(Prefix=cloud_path.key).limit(1)
```
jayqi, line 106: I'm thinking this might be more straightforward if it just returned the file object, or `None` if it doesn't exist. 

```
return next((obj for obj in self.s3.Bucket(cloud_path.bucket).objects.filter(Prefix=cloud_path.key).limit(1)), None)
```

instead of all of the functions that call this one having to iterate over it to get what they need.

cloudpathlib/s3/s3client.py
pjbull: Good point, there is a correctness/speed tradeoff here. I think this only goes wrong if a dir becomes a file or a file becomes a dir. I _think_ that's a pretty unlikely edge case. I'd be willing to skip a network call with that assumption, but could easily be convinced the network call is worth it for correctness. Thoughts?

cloudpathlib/s3/s3client.py
pjbull: :+1: Nice, will do

cloudpathlib/s3/s3client.py
jayqi: I think another scenario, based on how these functions are called:

- path exists either has a file or dir
- path gets cached locally
- cloud file(s) gets deleted through independent process
- `cp.exists()` will return `False` appropriately, but whichever `cp.is_dir()` and `cp.is_file()` was true before will still return `True` even though file no longer exists

pjbull: Yeah, files getting removed from the server could also cause unexpected behavior. I'll remove this optimization for now since we don't use this for existence anymore and we don't do file/dir checks in a lot of the cache management code so gains of not making the network call are probably pretty small.

jayqi (approved):
LGTM! 🎉

PR #121, opened 2021-02-13
Append filename to destination if source is file

### LINKED ISSUE #74: CloudPath.download_to doesn't work if src is file and destination is directory
If the source if a file and destination is a directory, we should make the `_download_file` destination into `destination / self.name`.

From the traceback:
```
~/code/cloudpathlib/cloudpathlib/cloudpath.py in download_to(self, destination)
    533         destination = Path(destination)
    534         if self.is_file():
--> 535             self.backend._download_file(self, destination)
    536         else:
    537             destination.mkdir(exist_ok=True)
```

genziano:
Fixes #74.

The `CloudPath.download_to` method adds the filename to the destination path if the cloud source is a file and the destination is a folder.

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -577,7 +577,7 @@ class CloudPath(metaclass=CloudPathMeta):
         destination = Path(destination)
         if self.is_file():
             if destination.is_dir():
-                destination = destination.joinpath(self.name)
+                destination = destination / self.name
             self.client._download_file(self, destination)
         else:
             destination.mkdir(exist_ok=True)
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -576,6 +576,8 @@ class CloudPath(metaclass=CloudPathMeta):
     def download_to(self, destination: Union[str, os.PathLike]):
         destination = Path(destination)
         if self.is_file():
+            if destination.is_dir():
+                destination = destination.joinpath(self.name)
             self.client._download_file(self, destination)
         else:
             destination.mkdir(exist_ok=True)
--- a/tests/test_cloudpath_file_io.py
+++ b/tests/test_cloudpath_file_io.py
@@ -85,6 +85,9 @@ def test_file_read_writes(rig, tmp_path):
 
     dl_dir = tmp_path / "directory"
     dl_dir.mkdir(parents=True, exist_ok=True)
+    assert not dl_dir.joinpath(p.name).exists()
+    p.download_to(dl_dir)
+    assert dl_dir.joinpath(p.name).exists() & dl_dir.joinpath(p.name).is_file()
     p3.download_to(dl_dir)
     cloud_rel_paths = sorted(
         # CloudPath("prefix://drive/dir/file.txt")._no_prefix_no_drive = "/dir/file.txt"
--- a/tests/test_cloudpath_file_io.py
+++ b/tests/test_cloudpath_file_io.py
@@ -100,9 +100,7 @@ def test_cloud_path_download_to(rig, tmp_path):
     p = rig.create_cloud_path("dir_0/file0_0.txt")
     dl_dir = tmp_path
     assert not (dl_dir / p.name).exists()
-    print(list(dl_dir.iterdir()))
     p.download_to(dl_dir)
-    print(list(dl_dir.iterdir()))
     assert (dl_dir / p.name).is_file()
 
 
--- a/tests/test_cloudpath_file_io.py
+++ b/tests/test_cloudpath_file_io.py
@@ -96,6 +96,16 @@ def test_file_read_writes(rig, tmp_path):
     assert cloud_rel_paths == dled_rel_paths
 
 
+def test_cloud_path_download_to(rig, tmp_path):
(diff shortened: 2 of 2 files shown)


jayqi:
The change itself looks good! For the tests, I would recommend adding a new test instead of adding to `test_file_read_writes`. `test_file_read_writes` is a bit big and unwieldy, and having a more targeted unit test is probably better for maintainability anyways.

jayqi:
By the way, if you write something like "Fixes #74" in your PR description, GitHub has special functionality to link this PR to that issue and will close the issue automatically when the PR is merged. If it's working, you'd see that the keyword has a dotted underline. The mechanics and triggering keywords are documented here: 

https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

genziano:
> By the way, if you write something like "Fixes #74" in your PR description, GitHub has special functionality to link this PR to that issue and will close the issue automatically when the PR is merged. If it's working, you'd see that the keyword has a dotted underline. The mechanics and triggering keywords are documented here:
> 
> https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

Nice one!

genziano:
> The change itself looks good! For the tests, I would recommend adding a new test instead of adding to `test_file_read_writes`. `test_file_read_writes` is a bit big and unwieldy, and having a more targeted unit test is probably better for maintainability anyways.

Good idea, I will separate that test. What do you think about having a class for testing, say `TestCloudPath`, where its methods implement tests for `CloudPath` methods?

jayqi:
> What do you think about having a class for testing, say TestCloudPath, where its methods implement tests for CloudPath methods?

For this project, we're using test modules (e.g., `test_cloudpath_file_io`) and descriptive test names for organizing our tests. From that regard, I don't think it's necessary to add test classes as another layer of organization. For parameterizing tests for the different types of cloud paths and handling setup/teardown, we're using fixtures—specifically the `rig` fixture. You can see how the rigs are set up on `conftest.py`. So, unless you have a compelling idea for how a test class could further simplify our testing code, I'm not in favor of using test classes.

genziano:
> > What do you think about having a class for testing, say TestCloudPath, where its methods implement tests for CloudPath methods?
> 
> For this project, we're using test modules (e.g., `test_cloudpath_file_io`) and descriptive test names for organizing our tests. From that regard, I don't think it's necessary to add test classes as another layer of organization. For parameterizing tests for the different types of cloud paths and handling setup/teardown, we're using fixtures—specifically the `rig` fixture. You can see how the rigs are set up on `conftest.py`. So, unless you have a compelling idea for how a test class could further simplify our testing code, I'm not in favor of using test classes.

I never used test classes myself, but I was curious to hear if you thought it could be a good idea. 

I see that most live tests are failing, does it have anything to do with the fact that this is a PR from a forked repository, for which there are no credentials?

cloudpathlib/cloudpath.py
```diff
@@ -576,6 +576,8 @@ def read_text(self):
     def download_to(self, destination: Union[str, os.PathLike]):
         destination = Path(destination)
         if self.is_file():
+            if destination.is_dir():
+                destination = destination.joinpath(self.name)
```
jayqi, line 580:
```suggestion
                destination = destination / self.name
```

Rather than `joinpath`, we prefer the `/` operator style of using pathlib.

tests/test_cloudpath_file_io.py
```diff
@@ -96,6 +96,16 @@ def test_file_read_writes(rig, tmp_path):
     assert cloud_rel_paths == dled_rel_paths
 
 
+def test_cloud_path_download_to(rig, tmp_path):
+    p = rig.create_cloud_path("dir_0/file0_0.txt")
+    dl_dir = tmp_path
+    assert not dl_dir.joinpath(p.name).exists()
```
jayqi, line 102:
```suggestion
    assert not (dl_dir / p.name).exists()
```

tests/test_cloudpath_file_io.py
```diff
@@ -96,6 +96,16 @@ def test_file_read_writes(rig, tmp_path):
...
+    print(list(dl_dir.iterdir()))
+    p.download_to(dl_dir)
+    print(list(dl_dir.iterdir()))
+    assert dl_dir.joinpath(p.name).exists() & dl_dir.joinpath(p.name).is_file()
```
jayqi, line 106:
```suggestion
    assert (dl_dir / p.name).is_file()
```

`is_file` will evaluate to False if the path doesn't exist.

jayqi (changes requested):
A few requested revisions.

> I see that most live tests are failing, does it have anything to do with the fact that this is a PR from a forked repository, for which there are no credentials?

Yes, that's the case. Sorry for the inconvenience—I'll think about how we can make this process better. What we did for a previous PR was for one of us (the maintainers) to copy the fork's branch over to the main repository so that the CI can run with the necessary credentials. For now, let's continue development in this PR until we're happy with the changes (using the CI steps before the live tests as a quality check). When it's ready, we can do the same approach of copying the branch to the main repo.

PR #122, opened 2021-02-15
[Staged] Append filename to destination if source is file

jayqi:
Staged version of #121 by @genziano on the main repo so we can run full test suite, including live tests with credentials.


pjbull (approved):
Looks great, thanks @genziano!

Issue #108, opened 2020-11-15
Add copy and copytree methods


jayqi:
We should have `copy` and `copytree` methods for copying from one cloud path to another, similar to how [shutil's functions](https://docs.python.org/3/library/shutil.html#shutil.copy) work.

genziano:
Hello, I was building a library for the same purpose, and I discovered this very nice one. I would like to contribute; is it ok if I pick this one up?

jayqi:
Yes! Thank you for your interest and the offer to contribute! Let us know if you have any questions.

pjbull:
That @genziano, that would be awesome. Just as a heads up, I think we may need two code paths:

(1) that uses the `_move_file` method on the client, which does a copy without transiting through the machine using cloudpath lib. (e.g., here is the [implementaiton for S3](https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/s3/s3client.py#L149-L164)). This is only guaranteed to work if the `client` on the `CloudPath` is the same client (since then the creds can access both buckets).

(2) that can work across cloud providers. This will likely need to do a download then upload and transit through the machine running cloudpathlib code.

genziano:
Great, I will work on it! And thanks for the tips.

I have one initial question: to which object should the methods `copy` and `copytree` belong? Or should they be functions in a submodule that replicates the `shutil` API?

pjbull:
IMO on `CloudPath` makes the most sense. Usage would look like. @jayqi did you have other thoughts?

```python
p1 = CloudPath("s3://b/file1.txt")

# copies the file from p1 to the passed path
p2 = p1.copy(CloudPath("s3://b/file2.txt"))

# str destination passes through to instantiation using the CloudPath factory on p1
p2 = p1.copy("s3://b/file2.txt")

# dir destination copies and keeps filename
p2 = p1.copy(CloudPath("s3://b/folder/"))

# copy on a dir raises error to use copytree is src is not a file
CloudPath("s3://b/folder/").copy("s3://b/folder2")

# copies the entire directory from folder1 to folder2
CloudPath("s3://b/folder/").copytree("s3://b/folder2")

```

machielg:
would be nice to have this feature, is there a plan for releasing?

pjbull:
This is being looked at in PR #142 now. @BenAsaf had mentioned he may be able to take a look at implementing in the near future.

PR #124, opened 2021-02-16
Split up CI workflows

jayqi:
This PR splits up the `tests` CI workflow, separating out the live tests into `tests-live` and the docs website preview into `docs-preview`. Those two things require credentials in the repo secrets, which aren't available to PRs from forks for reasonable security reasons. This way, we can run the mocked tests on the full build matrix and see if those builds pass, without the live tests or docs preview interfering. 

#123 shows a demo of this from a forked repo on my user account. 

### Alternative: Keep in one workflow, turn off fast-fail

An alternative to this approach is turning off fast-fail. The multiple workflow approach has the advantage of better visibility into the mocked tests success/fail states, as well as the more efficient compute use that comes from keeping fast-fail. 

### Bad Alternative: Make secrets available to forks

This is a bad idea because people can steal our credentials.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,25 +38,6 @@ jobs:
         run: |
           make test
 
-      - name: Set up Cloud SDK
-        uses: google-github-actions/setup-gcloud@master
-        with:
-          project_id: ${{ secrets.GCP_PROJECT_ID }}
-          service_account_key: ${{ secrets.GCP_SA_KEY }}
-          export_default_credentials: true
-
-      - name: Run live tests
-        run: |
-          make test-live-cloud
-        env:
-          LIVE_AZURE_CONTAINER: ${{ secrets.LIVE_AZURE_CONTAINER }}
-          AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}
-          LIVE_GS_BUCKET: ${{ secrets.LIVE_GS_BUCKET }}
-          LIVE_S3_BUCKET: ${{ secrets.LIVE_S3_BUCKET }}
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -38,59 +38,16 @@ jobs:
...
-          enable-commit-comment: false
-          overwrites-pull-request-comment: true
-          alias: deploy-preview-${{ github.event.number }}
-        env:
-          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
-          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
-        timeout-minutes: 1
-
       - name: Upload coverage to codecov
+        if: matrix.os == 'ubuntu-latest'
```
pjbull, line 49: This used to only upload on `3.8` as well—do we need to keep that check?

pjbull (approved):

.github/workflows/tests.yml
jayqi: I've been deliberately loosening this on some projects recently. Codecov will merge the reports together, which means you get better actual coverage on branching code between Python versions, e.g., new things they added to the standard library that older versions need a backport for.

pjbull: 🎉  TIL, cool!

jayqi:
Hi @genziano, wanted to let you know in case you're still working on #108 that we merged this PR earlier today. It splits out the CI steps that need credentials (like the live tests) into separate workflows, which should make the CI pass/fails more useful to you when you open a PR. Please rebase on the master branch if you haven't already.

genziano:
Hi @jayqi, thanks! I will rebase my branch and I will keep working on it.

genziano:
It seems that also the step "Update coverage to codecov" requires authentication, and it fails in the forked repository. Is there a way to set up codecov to run on my side as well, with my own account?

jayqi:
I had meant that if you open a PR on this repository from your fork, most of the CI workflows should run through without problem. 

Codecov should not require credentials for public repositories; if you open a PR on this repo, the CI here should push your PR's coverage results to _our_ Codecov account for this repository. It makes sense that your fork running the Codecov CI action wouldn't be able to directly push results to our Codecov though.

PR #125, opened 2021-02-26
Cloud implementation of shutil.copy and shutil.copytree

### LINKED ISSUE #58: API for creating a cloud file by uploading it
We have `download_to` to make it easy to copy a file from the cloud to the local filesystem. We don't have an equivalent `upload` or `upload_from` or `copy_from`  (naming is hard) method to easily create a cloud file from a local one.

E.g

```python
# path that does not exist yet
p = CloudPath("az://container/dir/new_file.txt")

#  copies "my_local_text_file.txt" to the path specified by p
p.upload("my_local_text_file.txt")
```

### LINKED ISSUE #108: Add copy and copytree methods

genziano:
Closes #108 and closes #58.

Work in progress.

Three methods of `CloudPath` implemented:
- `upload_from`, to upload a local file to a cloud file path;
- `copy`, to copy a cloud file to a cloud file or directory path;
- `copytree`, to copy a cloud directory path to another cloud directory path recursively.

Still to implement `copy` differently, which avoids download / upload when the client is the same.

(I will take care of the tests after we're set with the changes.)


jayqi:
A few thoughts:

- I'm not convinced about the shutil-style functions. The multiple cases could either be considered as more convenient or more confusing. I'm leaning that it's more confusing.
  - For cloud-to-local, we already `CloudPath.download_to` which feels more explicit and clear to me. 
  - For local-to-cloud, I think we should have an `upload_from` method which is the subject of #58 
  - Local-to-local that just dispatches to `shutil` feels unnecessary
- Then for the cloud-to-cloud, there are actually two cases, which Peter had mentioned in the issue:
  - We download the file and reupload it, like what you have.
  - If we're within the same cloud provider, we could copy directly without needing to download. I think the way you would check for that case is if the two `CloudPath` instances share the same client in `self.client`.

genziano:
> A few thoughts:
> 
> * I'm not convinced about the shutil-style functions. The multiple cases could either be considered as more convenient or more confusing. I'm leaning that it's more confusing.
>   
>   * For cloud-to-local, we already `CloudPath.download_to` which feels more explicit and clear to me.
>   * For local-to-cloud, I think we should have an `upload_from` method which is the subject of #58
>   * Local-to-local that just dispatches to `shutil` feels unnecessary
> * Then for the cloud-to-cloud, there are actually two cases, which Peter had mentioned in [this comment](https://github.com/drivendataorg/cloudpathlib/issues/108#issuecomment-775531767):
>   
>   * We download the file and reupload it, like what you have.
>   * If we're within the same cloud provider, we could copy directly without needing to download. I think the way you would check for that case is if the two `CloudPath` instances share the same client in `self.client`.

I will move the implementation as methods of `CloudPath`, also because I couldn't do what I thought because of circular imports. 

For the cloud-to-cloud, I was planning indeed to implement differently the case with the same client. And about this, do you have a preferred approach of checking whether the client of `src` and `dst` is the same? Checking equality on object level might be too strict.

jayqi:
> For the cloud-to-cloud, I was planning indeed to implement differently the case with the same client. And about this, do you have a preferred approach of checking whether the client of src and dst is the same? Checking equality on object level might be too strict.

I think actually that checking for the same instance is the right way to go, i.e., `self.client is dst.client`. cloudpathlib is implemented in a way that cloud path instances should share references to the same client instance unless explicitly created under a different one.

cloudpathlib/cloudshutil.py
```diff
@@ -0,0 +1,68 @@
...
+    src : PathLike
+        Source file.
+    dst : PathLike
+        Destination file or folder. If folder, the source file will be copied with its name.
+    follow_symlinks: bool, default = True
+        Passed only if both source and destination are local, in which case shutil.copy is used.
+    """
+    if not src.is_file():
+        raise ValueError('src must be a file.')
+    if isinstance(src, CloudPath) & isinstance(dst, CloudPath):
```
pjbull, line 37: One little note that all these should be `and` not `&`

cloudpathlib/cloudpath.py
```diff
@@ -595,6 +595,44 @@ def rmtree(self):
...
+
+    def copy(self, destination: "CloudPath"):
+        """ Copy self to destination folder of file, if self is a file. """
+        if not self.is_file():
+            raise ValueError(
+                f"Path {self} should be a file. To copy a directory tree use the method copytree."
+            )
+        temp_file = self._local
+        temp_file.parent.mkdir(parents=True)
+        self.download_to(temp_file)
```
pjbull, line 616: Instead of download to a `temp_file`, you can do `self.fspath` which will use the cached version if it is downloaded and download it to the cache if not. So the lines below can just use that. For example `destination.upload_from(self.fspath)`

cloudpathlib/cloudpath.py
```diff
@@ -595,6 +595,44 @@ def rmtree(self):
...
+        else:
+            (destination / self.name).upload_from(temp_file)
+
+    def copytree(self, destination: "CloudPath"):
+        """ Copy self to a directory, if self is a directory. """
+        if not self.is_dir():
+            raise ValueError(
+                f"Origin path {self} must be a directory. To copy a single file use the method copy."
+            )
+        if not destination.is_dir():
```
pjbull, line 628: This check might require that `destination` already exists, but I don't think that should be a requirement. Maybe  use `destination.exists() and destination.is_file()` instead?

pjbull:
Hi @genziano, seems like this is heading in the right direction—excited to get it merged!

Anything that we can help with in terms of implementation questions or writing tests?

cloudpathlib/cloudpath.py
jayqi: Agreed. `is_dir` will return `False` if `destination` does not exist. And, `shutil.copytree` will create the destination directory, so we should mirror that behavior.

genziano:
> Hi @genziano, seems like this is heading in the right direction—excited to get it merged!
> 
> Anything that we can help with in terms of implementation questions or writing tests?

Hi @pjbull! I have unfortunately been busy these last weeks, but next week I will work on finishing it up, and hopefully picking up some new issue.

Unrelated to this PR, I have a question: do you guys have some examples of usage with Azure blob storage? I'm not sure about how to create paths pointing to azure blob accounts. I'd also be happy to include some examples if needed!

pjbull:
> do you guys have some examples of usage with Azure blob storage

Is something like this what you mean or also including the authentication/credential flow (which is [in the docs](https://cloudpathlib.drivendata.org/api-reference/azblobclient/))

```python
from cloudpathlib import CloudPath, AzureBlobPath

# generic version starting with az:// is dispatched to AzureBlobPath
az_path = CloudPath("az://container/directory/file.txt")

# explicit version
az_path = AzureBlobPath("az://container/directory/file.txt")
```

genziano:
> > do you guys have some examples of usage with Azure blob storage
> 
> Is something like this what you mean or also including the authentication/credential flow (which is [in the docs](https://cloudpathlib.drivendata.org/api-reference/azblobclient/))
> 
> ```python
> from cloudpathlib import CloudPath, AzureBlobPath
> 
> # generic version starting with az:// is dispatched to AzureBlobPath
> az_path = CloudPath("az://container/directory/file.txt")
> 
> # explicit version
> az_path = AzureBlobPath("az://container/directory/file.txt")
> ```

Yes thanks, exactly that. I didn't figure out that the account name was just part of the connection string.

BenAsaf:
Hello!
Any news on this? I am looking forward to this PR :)
Can I assist in anyway?

PR #129, opened 2021-03-06
Rich comparison operators, hashing fix

jayqi:
- Adds rich comparison operators, which enables use with `sorted`
- Fixes a bug with `__hash__`. This was previously hashing the bound method `__repr__` (and not the string output of `__repr__`) and so gave a unique hash for every instance of a cloud path.


pjbull (approved):
Nice!

PR #130, opened 2021-03-07
Pydantic integration and AnyPath polymorphic class

### LINKED ISSUE #77: Add `AnyPath` class that will dispatch to either local or cloud
Just was working on some data pipeline code where it would be useful to be able to seamlessly pass either a cloud or a local path and have things just get handled happily.  

For example, we could replace code that looks like this
```python
DATA_ROOT = os.getenv("DATA_ROOT")

if DATA_ROOT.startswith("s3"):
    DATA_ROOT = S3Path(DATA_ROOT)
else:
    DATA_ROOT = Path(DATA_ROOT)
```

with code that look like this:
```python
DATA_ROOT = AnyPath(os.getenv("DATA_ROOT"))
```

### LINKED ISSUE #126: Add pydantic validator support
If we define a couple of class methods on `CloudPath`, then we can support the direct use of a cloud path class as a declared type on a Pydantic model.

https://pydantic-docs.helpmanual.io/usage/types/#custom-data-types

jayqi:
- Adds `AnyPath` polymorphic class. Closes #77. 
  - This class' constructor will instantiate a cloud path instance or a `pathlib.Path` instance appropriately based on the input.
  - This class also overloads `isinstance` and `issubclass` logic so that it's a virtual superclass of `CloudPath` and `Path`. 
  - New doc page: https://deploy-preview-130--gallant-agnesi-5f7bb3.netlify.app/anypath-polymorphism/
- Adds integration with Pydantic, so that any cloud path class or `AnyPath` can be used as a type declaration on a Pydantic model, and the validation will appropriately try to cast the input into the class. Closes #126. 
  - New doc page: https://deploy-preview-130--gallant-agnesi-5f7bb3.netlify.app/integrations/
- Refactor the way we specify extras so that it's all compactly in one file. 

Question:
- For now, I've just added pydantic to `requirements-dev.txt` for the tests. Should we make this an `extras` option instead? There's no direct runtime dependency by cloudpathlib on pydantic. You'd only actually need pydantic if you try to make a model and need to import `BaseModel`. 

Note:
- mypy has its own subtype-checking logic that does not use `issubclass`, so the `AnyPath` implementation doesn't quite pass mypy. I'm currently just putting in some `# type: ignore`s, but one may consider this implementation not-type-safe shenanigans. I may open an issue on mypy later to ask about if there is a recommended way to do things.


pjbull:
>For now, I've just added pydantic to requirements-dev.txt for the tests. Should we make this an extras option instead? There's no direct runtime dependency by cloudpathlib on pydantic. You'd only actually need pydantic if you try to make a model and need to import BaseModel.

Agreed to leave it out of the explicit requirements and just use it in dev.

docs/docs/anypath-polymorphism.md
```diff
@@ -0,0 +1,23 @@
+# AnyPath Polymorphic Class
+
+`cloudpathlib` implements a special `AnyClass` polymorphic class. This class will automatically instantiate a cloud path instance or a `pathlib.Path` instance appropriately from your input. It's also a virtual superclass of `CloudPath` and `Path`, so `isinstance` and `issubclass` checks will work in the expected way.
```
pjbull, line 3: typo s/`AnyClass`/`AnyPath`

docs/docs/anypath-polymorphism.md
```diff
@@ -0,0 +1,23 @@
...
+
+This functionality can be handy for situations when you want to support both local filepaths and cloud storage filepaths. If you use `AnyPath`, your code can switch between them seamlessly based on the contents of provided filepaths with needing any `if`-`else` conditional blocks.
+
```
pjbull, line 6: Should we add a note about how it does the dispatch?

pjbull (commented):
This is 🔥 🔥 🔥 

One thought is I don't think this implementation will support `Pure` paths. I'm fine opening an issue and deciding how we want to handle that later, but just wanted to mention.

jayqi:
I thought of the `PurePath` thing too, and I decided it makes sense not to support it, at least right now.
- We don't have any `PureCloudPath` classes. IIRC, use of cloudpathlib without authentication is undocumented and not explicitly supported. 
- I don't think we can automatically distinguish between inputs to `PurePath` and `Path` for dispatch. We'd have to provide more information and we'd have to invent an interface for it. 
- Most practical use cases that need to switch between cloud paths and local file paths probably care about interacting with real files and not a hypothetical filesystem.

docs/docs/anypath-polymorphism.md
jayqi: Added a section "How It Works"

pjbull (approved):

PR #131, opened 2021-03-07
Exceptions

jayqi:
- Exceptions:
    - Changed all custom `cloudpathlib` exceptions to be located in new `cloudpathlib.exceptions` module.
    - Changed all custom `cloudpathlib` exceptions to subclass from new base `CloudPathException`. This allows for easy catching of any custom exception from `cloudpathlib`.
    - Changed all custom exceptions names to end with `Error` as recommended by [PEP 8](https://www.python.org/dev/peps/pep-0008/#exception-names).
    - Changed various functions to throw new `CloudPathFileExistsError`, `CloudPathIsADirectoryError` or `CloudPathNotADirectoryError` exceptions instead of a generic `ValueError`.
    - Removed exception exports from the root `cloudpathlib` package namespace. Import from `cloudpathlib.exceptions` instead if needed.
- Fixed `AzureBlobClient` instantiation to throw new error `MissingCredentialsError` when no credentials are provided, instead of `AttributeError`. `LocalAzureBlobClient` changed to throw error under same conditions accordingly.
- Fixed `GSClient` to instantiate as anonymous with public access only when instantiated with no credentials, instead of erroring.

Some additional testing updates:
- Change `conftest.py` only run `load_dotenv` if `USE_LIVE_CLOUD` is set to 1. This was fooling my testing because an unset env var was present in my `.env` file. 
- Tests for specific cloud provider implementations have been parametrized to also test their `cloudpathlib.local` equivalents.


tests/conftest.py
```diff
@@ -29,7 +29,8 @@
 from .mock_clients.mock_s3 import mocked_session_class_factory
 
 
-load_dotenv(find_dotenv())
+if os.getenv("USE_LIVE_CLOUD") == "1":
+    load_dotenv(find_dotenv())
```
jayqi, line 33: This had fooled my local testing by loading environment variables when I wasn't expecting them for the mocked tests.

jayqi:
Ended up doing more on fixing tests than I expected, but this is actually ready now @pjbull

pjbull (approved):
This looks good to me, thanks for digging in and making this neat and tidy. Loving the readable, standardized exceptions

PR #135, opened 2021-03-15
Add pandas-path integration

pjbull:
- Adds `.cloud` accessor to `pd.Series` and `pd.Index` via a pandas-path [custom accessor](https://github.com/drivendataorg/pandas-path/#custom-path-accessors)
 - Adds simple tests for the integration
 - Adds docs for the integration
 - Updates changelog

### DIFF
--- a/docs/docs/integrations.md
+++ b/docs/docs/integrations.md
@@ -34,5 +34,68 @@ fancy2.path
 #> PosixPath('mydir/myfile.txt')
 ```
 
+
+## Pandas and [pandas-path](https://github.com/drivendataorg/pandas-path)
+
+[`pandas-path`](https://github.com/drivendataorg/pandas-path) provides `pathlib` functions through as custom accessor (`.path`). The library also supports registering custom accessors for any class that implements that pathlib API.
+
+### The `.cloud` accessor
+
+We expose a `.cloud` accessor on `pandas.Series` and `pandas.Index` objects if you import it. This allows you to access any `CloudPath` method or property directly on a Series by just adding `.cloud`.
+
+To use the `.cloud` accessor, you must have `pandas-path` installed through pip:
+
+```
+pip install pandas-path
+```
+
+All you need to do to register the accessor is `from cloudpathlib.pandas import cloud`.
+
+For example:
+
+```python
+from cloudpathlib.pandas import cloud
+import pandas as pd
+
+pd.Series([
+    's3://cats/1.jpg',
+    's3://cats/2.jpg',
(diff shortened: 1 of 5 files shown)


requirements-dev.txt
```diff
@@ -12,6 +12,7 @@ mkdocs-material>=7
 mkdocstrings>=0.15
 mypy
 pandas
+pandas-path>=0.3.0
```
jayqi, line 15: Wonder if it makes sense to add a `pandas` extras option.

docs/docs/integrations.md
```diff
@@ -34,5 +34,68 @@ fancy2.path
...
+
+We expose a `.cloud` accessor on `pandas.Series` and `pandas.Index` objects if you import it. This allows you to access any `CloudPath` method or property directly on a Series by just adding `.cloud`.
+
+To use the `.cloud` accessor, you must have `pandas-path` installed through pip:
+
+```
+pip install pandas-path
+```
+
+All you need to do to register the accessor is `from cloudpathlib.pandas import cloud`.
```
jayqi, line 52: Why the extra module layer, instead of just `import cloudpathlib.pandas`?

requirements-dev.txt
pjbull: I think it's kind of conceptually nice to have the extras be focused on cloud providers since that's the major differentiator.

docs/docs/integrations.md
pjbull: This is to mirror the the pandas-path import which looks like `from pandas_path import path`. I think it's nice to (1) make the structure similar, and (2) see the name of the accessor when you do the import so you can see how they might get tied together.

pjbull:
@jayqi Let me know if you have thoughts on putting this functionality in cloudpathlib. Now that it is implemented, I could also see this making sense in the `pandas-path` package rather than here. I don't feel strongly either way.

jayqi:
Yeah, I guess it could go either way:

---

```
pip install cloudpathlib[s3]
pip install pandas-path
```

```
from cloudpathlib.pandas import cloud
```

---

```
pip install pandas-path[s3]
```

```
from pandas_path import cloud
```

Now that I've seen it, I guess the interface looks cleaner to have this live in `pandas-path`.

jayqi:
@pjbull the more I've thought about this one, the more that I think having this as an extension within `pandas_path` makes more sense than including it in `cloudpathlib`.

pjbull:
Stale, closing

Issue #136, opened 2021-03-18
How to use cloudpathlib for accessing custom S3 installations? 


YevheniiSemendiak:
For instance, I've deployed MinIO to my cluster and have its endpoint URL.
The only solution that comes to my mind is to instantiate CloudPath with a boto3 client, authenticated to this endpoint.
Am I correct with this assumption?

jayqi:
I'm not familiar with MinIO. Is this documentation describing something similar your situation and how you'd access your S3 with boto3? https://docs.min.io/docs/how-to-use-aws-sdk-for-python-with-minio-server.html

YevheniiSemendiak:
@jayqi just posted a PR #137 as an example for it

PR #137, opened 2021-03-18
[WIP] Add possibility to parameterize S3 service URL

YevheniiSemendiak:
Just a minor thing, related to #136
I didn't found contribution guides, so this is potentially draft PR.
Nevertheless, WDYT about the possibility to support non-AWS S3 deployments?

Re the PR: what else should be added?

P.S. Thank you folks for an awesome lib 👍

### DIFF
--- a/docs/docs/authentication.md
+++ b/docs/docs/authentication.md
@@ -69,10 +69,15 @@ provinding http/https URL including port.
 ```python
 from cloudpathlib import S3Client, CloudPath
 
+# create a client pointing to the endpoint
 client = S3Client(endpoint_url="http://my.s3.server:1234")
-# set this client as a default one
-client.set_as_default_client()
+# option 1: use the client to create paths
 cp1 = client.CloudPath("s3://cloudpathlib-test-bucket/")
-# or pass the client as keyword argument
+
+# option 2: pass the client as keyword argument
 cp2 = CloudPath("s3://cloudpathlib-test-bucket/", client=client)
+
+# option3: set this client as the default so it is used in any future paths
+client.set_as_default_client()
+cp3 = CloudPath("s3://cloudpathlib-test-bucket/")
 ```
--- a/docs/docs/authentication.md
+++ b/docs/docs/authentication.md
@@ -32,7 +32,7 @@ cloud_path.client
 
 All subsequent instances of that service's cloud paths (in the example, all subsequent `S3Path` instances) will reference the same client instance.
 
-You can also explicitly instantiate a client instance. You will need to do so if you want to authenticate using any option other than the environment variables from the table in the previous section. (To see what those options are, check out the API documentation pages linked to in the table above.) You can then use that client instance's cloud path factory method, or pass it into a cloud path instantiation
+You can also explicitly instantiate a client instance. You will need to do so if you want to authenticate using any option other than the environment variables from the table in the previous section. (To see what those options are, check out the API documentation pages linked to in the table above.) You can then use that client instance's cloud path factory method, or pass it into a cloud path instantiation.
 
 ```python
 from cloudpathlib import S3Client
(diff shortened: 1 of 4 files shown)


pjbull:
Happy to take this change! A few questions:
 - Can we add some notes to the docs about Min.IO and other S3 compatible object stores so that people can find it? Maybe at the end [of the authentication path](https://cloudpathlib.drivendata.org/authentication/)
 - Can we make a similar note in the docstring for the endpoint param?
 - I don't think we need additional mocked tests, but we may want a live server config to do this. Is there a min.io test server or hosted version with a free tier we could get creds for the test suite to access?

YevheniiSemendiak:
The notes are added here: 258392b
And the custom S3 rig here: 02f7a8c
Hope I get what you meant properly.

docs/docs/authentication.md
```diff
@@ -59,3 +59,20 @@ If you need a reference to the default client:
...
+```python
+from cloudpathlib import S3Client, CloudPath
+
+client = S3Client(endpoint_url="http://my.s3.server:1234")
+# set this client as a default one
+client.set_as_default_client()
+cp1 = client.CloudPath("s3://cloudpathlib-test-bucket/")
+# or pass the client as keyword argument
+cp2 = CloudPath("s3://cloudpathlib-test-bucket/", client=client)
+```
```
pjbull, line 78: I might set this up to show three scenarios:

```python
from cloudpathlib import S3Client, CloudPath

# create a client pointing to the endpoint
client = S3Client(endpoint_url="http://my.s3.server:1234")

# option 1: use the client to create paths
cp1 = client.CloudPath("s3://cloudpathlib-test-bucket/")

# option 2: pass the client as keyword argument
cp2 = CloudPath("s3://cloudpathlib-test-bucket/", client=client)

# option3: set this client as the default so it is used in any future paths
cp3 = CloudPath("s3://cloudpathlib-test-bucket/")
```

tests/conftest.py
```diff
@@ -212,6 +219,49 @@ def s3_rig(request, monkeypatch, assets_dir):
         bucket.objects.filter(Prefix=test_dir).delete()
 
 
+@fixture()
+def custom_s3_rig(request, monkeypatch, assets_dir):
```
pjbull, line 223: Let's add a docstring that this is used for testing things like MinIO/ceph?

tests/conftest.py
```diff
@@ -212,6 +219,49 @@ def s3_rig(request, monkeypatch, assets_dir):
...
+    drive = os.getenv("CUSTOM_S3_BUCKET", "bucket")
+    custom_endpoint_url = os.getenv("CUSTOM_S3_ENDPOINT")
+    custom_key_id = os.getenv("CUSTOM_S3_KEY_ID")
+    custom_secret_key = os.getenv("CUSTOM_S3_SECRET_KEY")
+    test_dir = create_test_dir_name(request)
+
+    # Upload test assets
```
pjbull, line 230: Let's keep the same `if os.getenv("USE_LIVE_CLOUD") == "1":` logic as in the S3 one so that local tests are fully mocked (`endpoint_url` is basically ignored in the mock), but we are running the integration tests against live backends we do this copying.

pjbull (changes requested):
Looking great @YevheniiSemendiak, thanks! Just a couple small things to tweak.

Once those are done, I'll probably move this over to a PR branch on the repo so that I can get the integration tests that use a live backend properly configured before merging to main.

tests/conftest.py
YevheniiSemendiak: done here fc113dc

tests/conftest.py
YevheniiSemendiak: done here fc113dc

docs/docs/authentication.md
YevheniiSemendiak: done here fc113dc

PR #138, opened 2021-03-23
Add possibility to parameterize S3 service URL

pjbull:
#137 added ability to specify the endpoint url for `S3Path`

That work is accepted, and we need to now test if the configuration of the live tests for that scenario work from a non-fork branch.

### DIFF
--- a/.github/workflows/tests-live.yml
+++ b/.github/workflows/tests-live.yml
@@ -43,7 +43,7 @@ jobs:
           LIVE_S3_BUCKET: ${{ secrets.LIVE_S3_BUCKET }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          CUSTOM_S3_BUCKET: ${{CUSTOM_S3_BUCKET}}
-          CUSTOM_S3_KEY_ID: ${{CUSTOM_S3_KEY_ID}}
-          CUSTOM_S3_SECRET_KEY: ${{CUSTOM_S3_SECRET_KEY}}
-          CUSTOM_S3_ENDPOINT: ${{CUSTOM_S3_ENDPOINT}}
+          CUSTOM_S3_BUCKET: ${{secrets.CUSTOM_S3_BUCKET}}
+          CUSTOM_S3_KEY_ID: ${{secrets.CUSTOM_S3_KEY_ID}}
+          CUSTOM_S3_SECRET_KEY: ${{secrets.CUSTOM_S3_SECRET_KEY}}
+          CUSTOM_S3_ENDPOINT: ${{secrets.CUSTOM_S3_ENDPOINT}}
--- a/.github/workflows/tests-live.yml
+++ b/.github/workflows/tests-live.yml
@@ -43,3 +43,7 @@ jobs:
           LIVE_S3_BUCKET: ${{ secrets.LIVE_S3_BUCKET }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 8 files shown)


docs/docs/authentication.md
```diff
@@ -59,3 +59,25 @@ If you need a reference to the default client:
...
+In such cases, the service endpoint will be different from the AWS object store endpoints (used by default).
+To specify a custom endpoint address, you will need to manually instantiate `Client` with the `endpoint_url` parameter,
+provinding http/https URL including port.
+
+```python
+from cloudpathlib import S3Client, CloudPath
+
+# create a client pointing to the endpoint
+client = S3Client(endpoint_url="http://my.s3.server:1234")
+# option 1: use the client to create paths
```
jayqi, line 74:
```suggestion
client = S3Client(endpoint_url="http://my.s3.server:1234")

# option 1: use the client to create paths
```

New line here for consistency.

tests/conftest.py
```diff
@@ -212,6 +219,63 @@ def s3_rig(request, monkeypatch, assets_dir):
...
+        )
+
+    rig = CloudProviderTestRig(
+        path_class=S3Path,
+        client_class=S3Client,
+        drive=drive,
+        test_dir=test_dir,
+        endpoint_url=custom_endpoint_url,
+        aws_access_key_id=custom_key_id,
+        aws_secret_access_key=custom_secret_key,
```
jayqi, line 268: I don't think adding `**client_kwargs` and passing these in here is a change we actually need to make to the test rig, because in 271 below we can just explicitly pass these in to the default client instantiation.

jayqi (changes requested):
Tests pass and I think this looks good overall, but I think there are some changes we can make. @pjbull do you want to fix this or would you like me to?

pjbull:
@jayqi FYI, the tests don't actually pass. (The rig wasn't added to the list and so they weren't run).

There are some failing tests that may actually be MinIO/S3 incompatibility, but I haven't gotten to the bottom of them yet.

pjbull:
Also, changes you suggest make sense, feel free to make those and look at the failing tests. I'll check with you next time I have a change to dig in to see if you uncovered anything

.github/workflows/tests-live.yml
```diff
@@ -43,3 +43,7 @@ jobs:
           LIVE_S3_BUCKET: ${{ secrets.LIVE_S3_BUCKET }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          CUSTOM_S3_BUCKET: ${{CUSTOM_S3_BUCKET}}
+          CUSTOM_S3_KEY_ID: ${{CUSTOM_S3_KEY_ID}}
+          CUSTOM_S3_SECRET_KEY: ${{CUSTOM_S3_SECRET_KEY}}
+          CUSTOM_S3_ENDPOINT: ${{CUSTOM_S3_ENDPOINT}}
```
pjbull, line 49: Do these need the same `secrets.`?

jayqi: Those are there already. I thought you added those? 😆

pjbull: I meant should `${{CUSTOM_S3_BUCKET}}` be `${{ secrets.CUSTOM_S3_BUCKET }}`

pjbull: but yes, i did add the secrets!

jayqi: Oh, you are totally right. That was confusing, it declared my workflow invalid, but didn't show it as a failure.

jayqi:
@pjbull I think this looks ready now if you have time to take a look.

tests/test_client.py
```diff
@@ -2,7 +2,10 @@
 
 
 def test_default_client_instantiation(rig):
-    rig.client_class._default_client = None
+    if not getattr(rig, "is_custom_s3", False):
+        # Skip resetting the default client for custom S3 endpoint, but keep the other tests,
+        # since they're still useful.
+        rig.client_class._default_client = None
```
pjbull, line 8: 👏 I hadn't figured out why this one was failing!

PR #139, opened 2021-03-26
Remove unnecessary .join()

isms:


jayqi (approved):
Thanks! Not sure why I had written it that way. Maybe I was originally doing `"\n".join()` or something.

Issue #128, opened 2021-03-05
open(cloud_path) acts on cached local file instead of cloud file


jayqi:
Unfortunately the following code does not behave as expected:

```python
with open(cloud_path, "w") as fp:
    fp.write("foo")
```

(as opposed to using `cloud_path.open("w")` which does work as expected). 

What will happen is that this will only modify the local cached copy of the file without uploading it to the cloud. Then further attempts to interact with the cloud path will correctly raise `OverwriteNewerLocal` exceptions. Because cloud paths have an `__fspath__` method that returns the local file path, the `open` function will only interact with that local file path. 

Users can still get the correct behavior by using the `open` method. However, I've seen a lot of people who are still new to regular pathlib using `open(path, "w")` on their `Path` objects, so I expect this may be a common pattern for users of cloudpathlib as well.

jayqi:
The gist of the problem here is that our `os.PathLike` support fails in write situations. 

Because `CloudPath` implements `__fspath__` and therefore the `os.PathLike` protocol, it can be accepted by anything that supports an `os.PathLike` input, which will call `__fspath__` and do whatever with the returned filepath. This fine and dandy for read operations, because the local cache file's path is returned and the caller will read that file, and we have the ability to refresh the cache before that happens. 

However, the problem happens when the caller writes. We're able to refresh the cache within `__fspath__` before returning the local cache's path, but after we return, we're out of that method's scope and can't do anything else. The caller will write to the local cache, but we won't be able to sync the cache back up to the cloud. 

This may be impossible to fix. Once `__fspath__` returns, we won't be able to do anything. We'd really like `__fspath__` to behave like a context manager so we can run some code after, but unfortunately we can't actually do that because the `os.PathLike` protocol requires `__fspath__` to return `bytes` or `str`. We'd need the Python standard library to implement a protocol that supports a context manager and then they have to convince all of the other libraries to support their new context manager interface. `os.PathLike` has been out since Python 3.6 and there are still lots of places that don't support it yet and still only support filepaths as strings. 

---

If it's the case that this is impossible to fix, then we have two options:

#### 1. Keep our `os.PathLike` support but try to detect when it's being used in a write case so we can raise an exception. 

@pjbull suggested that we could maybe use the `inspect` module to trace the callstack. We could potentially try to detect if it's being used by the built-in `open` function and then check whether `mode` is set to one of the writable ones. 

[...]

PR #140, opened 2021-04-03
[WIP] Raise exception when open with write mode in call stack

jayqi:
A pass at trying to detect when `__fspath__` is called by `open` with a write mode. IMO it doesn't work all that well. It appears `open` is a weird function that doesn't have its own frame in the call stack. It jumps from `__fspath__`'s frame directly to the frame that calls `open`. I don't think there's a straightforward way to get object references from the frames (it's bytecode or source), so also hard to reliably detect if `open` is being called on the cloud path instance, or to pick up on what mode `open` is being called with.

### DIFF
--- a/tests/test_cloudpath_file_io.py
+++ b/tests/test_cloudpath_file_io.py
@@ -5,7 +5,11 @@ from time import sleep
 
 import pytest
 
-from cloudpathlib.exceptions import CloudPathIsADirectoryError, DirectoryNotEmptyError
+from cloudpathlib.exceptions import (
+    BuiltInOpenWriteError,
+    CloudPathIsADirectoryError,
+    DirectoryNotEmptyError,
+)
 
 
 def test_file_discovery(rig):
@@ -113,3 +117,26 @@ def test_os_open(rig):
     p = rig.create_cloud_path("dir_0/file0_0.txt")
     with open(p, "r") as f:
         assert f.readable()
+
+    with pytest.raises(BuiltInOpenWriteError):
+        with open(p, "w") as f:
+            pass
+
+    with pytest.raises(BuiltInOpenWriteError):
+        with open(p, "wb") as f:
+            pass
+
+    with pytest.raises(BuiltInOpenWriteError):
+        with open(p, "a") as f:
+            pass
+
+    with pytest.raises(BuiltInOpenWriteError):
+        with open(p, "r+") as f:
+            pass
+
+    # with pytest.raises(BuiltInOpenWriteError):
+    #     with open(
+    #         p,
+    #         "w",
+    #     ) as f:
+    #         pass
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -4,11 +4,14 @@ import collections.abc
 import fnmatch
 import os
 from pathlib import Path, PosixPath, PurePosixPath, WindowsPath
+import inspect
+import traceback
 from typing import Any, IO, Iterable, Optional, TYPE_CHECKING, Union
 from urllib.parse import urlparse
 from warnings import warn
 
 from .exceptions import (
+    BuiltInOpenWriteError,
     ClientMismatchError,
     CloudPathFileExistsError,
     CloudPathIsADirectoryError,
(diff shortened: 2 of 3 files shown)


tests/test_cloudpath_file_io.py
```diff
@@ -113,3 +117,26 @@ def test_os_open(rig):
...
+    with pytest.raises(BuiltInOpenWriteError):
+        with open(p, "r+") as f:
+            pass
+
+    # with pytest.raises(BuiltInOpenWriteError):
+    #     with open(
+    #         p,
+    #         "w",
+    #     ) as f:
+    #         pass
```
jayqi, line 142: This case fails because the test frame's `code_context` looks like this: `['        with open(\n']`. Not sure what to do about it.

pjbull:
Ok @jayqi, this really sent me down a 🐰 🕳️ ...

Here's a notebook showing a few more options:
https://gist.github.com/pjbull/1b8f92b84a40cd6fe2033468f98594a4

Thoughts on those?

jayqi:
Wow, well done. 👏  It looks to me like you're capturing the main cases. I think grabbing the entire source of the frame and parsing the AST is probably the only reliable way to do this. One concern I have is how much overhead this adds to each call, and how it scales with the frame's size.

pjbull:
Yeah, that's exactly my concern too. I'll see if I can add some benchmarking. No clue how long those things take....

pjbull:
OK, I updated that gist above with some benchmarking. Here are my takeaways:
 - Adding the check makes the `__fspath__` take ~10x the time of having no check
 - On my machine, the slowdown is in the same ballpark as writing a 48KB file or reading a 1.6MB on. As file sizes grow larger than those, the slowdown will start to be overwhelmed by the cost of reading/writing.
 - `inspect.getsource` is the most expensive call. I didn't track down a faster call for this in some poking around.
 - Neither AST parsing nor regex change execution time appreciably, even for reasonably large code contexts, so choosing one of those should not be based on performance.

So, I think if we go with **Option 1** as outlined in #128  we should:
(1) add the AST version based on `getcurrentframe` which is fast and robust as compared to the other options, and (2) implement an environment variable to allow opting out of the performance hit if you are sure you don't use `open` with cloudpaths.

For me, options 2-4 as outlined in #128  are also still on the table given this research.

jayqi:
Do you think `inspect.getsource` might have a different speed if it's getting the source from an installed version of cloudpathlib? (Editable and normal)

6 ms wall time per run isn't _untenably_ slow but I guess not trivial. Adds a minute to 10,000 files.

pjbull:
> Do you think inspect.getsource might have a different speed if it's getting the source from an installed version of cloudpathlib? (Editable and normal)

I think since the frame we want to look at is from the caller's code, it will depend on how their code is installed so not sure that we can speed it up.

I did just poke at bytecode disassembly with [`dis`](https://docs.python.org/3/library/dis.html) as well, and that gave me the same order of magnitude as `getsource` (~5ms).

Agreed that 6ms isn't large enough to rule this approach out.

I think to improve the user friendliness of working with paths, I'm inclined towards this change.

pjbull:
@jayqi Check out the latest commits. This seems to work, but it does add a ton of complex code for a narrow-ish use case.

jayqi (commented):
"Pull request authors can't request changes on their own pull request." 😂  🙄

pjbull:
Another rev here. I thought that with the AST we lost the line numbers for matching writeable opens we detected and the entry point from  calls, but it turns out that with a little bit of tracking, we can use both the AST approach and raise the error on the right line.

Tests path locally on 3.8 on linux as well, so there is likely some fix needed for Python<3.8

Small bonus fix: silence warnings in tests by checking for attributes in `__del__` methods

jayqi:
Uh oh, did we end up using a Python 3.8-only feature? It looks like the 3.8 build passed but 3.6 and 3.7 failed.

pjbull:
Yeah, see above:
> Tests path locally on 3.8 on linux as well, so there is likely some fix needed for Python<3.8

I'm seeing differences in what line `inspect` reports for a given frame across versions. Need to dig in more to find a fix.

pjbull:
OK, I think this is ready for review. It is passing all the tests.

I do think that this is a pretty substantial amount of complexity to add for maitenance, so we may want to turn this off by default if we get lots of bug reports that come up in this code block.

I wouldn't be surprised if there are cases where the `inspect` calls don't do what we expect given all the different ways one can run Python.


### Performance update
----------------
Here's info on timing for this implementation:

Script:
```python
import statistics
import time


if __name__ == "__main__":
    p = CloudPath('s3://cloudpathlib-test-bucket/5bmg9iTs74wmewasLTP44W-test_cloudpath_file_io-test_file_discovery/dir_0/file0_0.txt')

    it = 100
    res = []

    for _ in range(it):
        start = time.time()

        with open(p, "r"):
            pass

        res.append(time.time() - start)

    print(f"Median time {it} iterations: {statistics.median(res)}")
```

Results:

**Around +0.05s on average for each call to `open`**

```
❯ CLOUDPATHLIB_CHECK_UNSAFE_OPEN=True python timing.py
Median time 100 iterations: 0.2683759927749634

❯ CLOUDPATHLIB_CHECK_UNSAFE_OPEN=False python timing.py
Median time 100 iterations: 0.2952016592025757
```

PR #142, opened 2021-04-21
Cloud implementation of shutil.copy and shutil.copytree

### LINKED ISSUE #58: API for creating a cloud file by uploading it

### LINKED ISSUE #108: Add copy and copytree methods

pjbull:
Seems like @genziano was busy (thanks for the initial work!), so I moved over his WIP from #125 to this branch.

@benasaf feel free to pick it up from here.

--------

Moved over PR from #125:

* Draft of cloudpathlib.copy, that replicates shutil.copy

* Import only PathLike from os

* Reuse _local_to_cloud_copy for cached files

* try to use the intended cached _local version

* Update cloudshutil.py

* Implemented the CloudPath methods: upload_from, copy, copytree

* Add docstrings

* Fix indentation

* fix unnecessary imports

* Fix typing issues

* Allow CloudPath.copy to a directory as well

Closes #58 
Closes #108

### DIFF
--- a/cloudpathlib/anypath.py
+++ b/cloudpathlib/anypath.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 from typing import Union
 
@@ -52,3 +53,14 @@ class AnyPath(metaclass=AnyPathMeta):
         https://pydantic-docs.helpmanual.io/usage/types/#custom-data-types"""
         # Note __new__ is static method and not a class method
         return cls.__new__(cls, value)
+
+
+def to_anypath(s: Union[str, os.PathLike]) -> Union[CloudPath, Path]:
+    """Convenience method to convert a str or os.PathLike to the
+    proper Path or CloudPath object using AnyPath.
+    """
+    # shortcut pathlike items that are already valid Path/CloudPath
+    if isinstance(s, (CloudPath, Path)):
+        return s
+
+    return AnyPath(s)  # type: ignore
--- a/cloudpathlib/client.py
+++ b/cloudpathlib/client.py
@@ -60,7 +60,7 @@ class Client(abc.ABC, Generic[BoundedCloudPath]):
     @abc.abstractmethod
     def _download_file(
         self, cloud_path: BoundedCloudPath, local_path: Union[str, os.PathLike]
-    ) -> Union[str, os.PathLike]:
+    ) -> Path:
         pass
 
     @abc.abstractmethod
--- a/cloudpathlib/client.py
+++ b/cloudpathlib/client.py
@@ -83,7 +83,9 @@ class Client(abc.ABC, Generic[BoundedCloudPath]):
         pass
 
     @abc.abstractmethod
-    def _move_file(self, src: BoundedCloudPath, dst: BoundedCloudPath) -> BoundedCloudPath:
+    def _move_file(
+        self, src: BoundedCloudPath, dst: BoundedCloudPath, remove_src: bool = True
+    ) -> BoundedCloudPath:
         pass
 
     @abc.abstractmethod
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -585,12 +585,12 @@ class CloudPath(metaclass=CloudPathMeta):
         return self._dispatch_to_local_cache_path("read_text")
 
     # ===========  public cloud methods, not in pathlib ===============
-    def download_to(self, destination: Union[str, os.PathLike]):
+    def download_to(self, destination: Union[str, os.PathLike]) -> os.PathLike:
         destination = Path(destination)
         if self.is_file():
             if destination.is_dir():
                 destination = destination / self.name
-            self.client._download_file(self, destination)
+            return self.client._download_file(self, destination)
         else:
             destination.mkdir(exist_ok=True)
             for f in self.iterdir():
@@ -601,6 +601,8 @@ class CloudPath(metaclass=CloudPathMeta):
                 rel_dest = str(f)[len(rel) :]
                 f.download_to(destination / rel_dest)
 
+            return destination
+
     def rmtree(self):
         """Delete an entire directory tree."""
         if self.is_file():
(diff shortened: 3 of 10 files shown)


genziano:
Hey @pjbull, indeed, sorry for leaving it unfinished but it's a very busy period! Glad to have contributed!

cloudpathlib/cloudpath.py
```diff
@@ -609,6 +611,90 @@ def rmtree(self):
...
+    def copytree(
+        self, destination: "CloudPath", force_overwrite_to_cloud: bool = False
+    ) -> Union[os.PathLike, "CloudPath"]:
+        """Copy self to a directory, if self is a directory."""
+        if not self.is_dir():
+            raise ValueError(
+                f"Origin path {self} must be a directory. To copy a single file use the method copy."
+            )
+        if destination.exists() and destination.is_file():
+            raise ValueError("Destination path {destination} of copytree must be a directory.")
```
jayqi, line 682: I think we want to use [`CloudPathFileExistsError`](https://cloudpathlib.drivendata.org/api-reference/exceptions/#cloudpathlib.exceptions.CloudPathFileExistsError).

cloudpathlib/cloudpath.py
```diff
@@ -609,6 +611,90 @@ def rmtree(self):
...
+                return (destination / self.name).upload_from(
+                    self.fspath, force_overwrite_to_cloud=force_overwrite_to_cloud
+                )
+
+    def copytree(
+        self, destination: "CloudPath", force_overwrite_to_cloud: bool = False
+    ) -> Union[os.PathLike, "CloudPath"]:
+        """Copy self to a directory, if self is a directory."""
+        if not self.is_dir():
+            raise ValueError(
```
jayqi, line 678: This should be [`CloudPathNotADirectoryError`](https://cloudpathlib.drivendata.org/api-reference/exceptions/#cloudpathlib.exceptions.CloudPathNotADirectoryError)

cloudpathlib/cloudpath.py
```diff
@@ -609,6 +611,90 @@ def rmtree(self):
             )
         self.client._remove(self)
 
+    def upload_from(
+        self, source: Union[str, os.PathLike], force_overwrite_to_cloud: bool = False
```
jayqi, line 615: It's not clear to me why we need `force_overwrite_to_cloud` if `source` can be an arbitrary file. I would expect this method to just always overwrite. Conceptually similar methods for moving or copying will just overwrite things. 

Relatedly, I think the checking logic for this should remain in `_upload_local_to_cloud` instead of the more general `_upload_file_to_cloud`.

cloudpathlib/client.py
```diff
@@ -83,7 +83,9 @@ def _list_dir(
         pass
 
     @abc.abstractmethod
-    def _move_file(self, src: BoundedCloudPath, dst: BoundedCloudPath) -> BoundedCloudPath:
+    def _move_file(
+        self, src: BoundedCloudPath, dst: BoundedCloudPath, remove_src: bool = True
```
jayqi, line 87: Personally, I'd find it more intuitive to have separate methods `_move_file` and `_copy_file`, where the base implementation of `_move_file` would be something like:

```python
def _move_file(self, src: BoundedCloudPath, dst: BoundedCloudPath) -> BoundedCloudPath:
    self._copy_file(src, dst)
    if src != dst:
        self._remove(src)
    return dst
```

jayqi (changes requested):

cloudpathlib/cloudpath.py
pjbull: Yeah, I guess that overwriting your canonical data in the cloud is potentially the worst thing you could do so we were being extra careful to be explicit here that you want to overwrite. I'm on the fence for this one as a default.

cloudpathlib/client.py
pjbull: Opened #143

cloudpathlib/cloudpath.py
```diff
@@ -609,6 +611,90 @@ def rmtree(self):
...
+                return destination.upload_from(
+                    self.fspath, force_overwrite_to_cloud=force_overwrite_to_cloud
+                )
+            else:
+                return (destination / self.name).upload_from(
+                    self.fspath, force_overwrite_to_cloud=force_overwrite_to_cloud
+                )
+
+    def copytree(
+        self, destination: "CloudPath", force_overwrite_to_cloud: bool = False
```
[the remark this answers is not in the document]
pjbull: Opened #145

cloudpathlib/anypath.py
```diff
@@ -52,3 +53,14 @@ def _validate(cls, value) -> Union[CloudPath, Path]:
...
+
+def to_anypath(s: Union[str, os.PathLike]) -> Union[CloudPath, Path]:
+    """Convenience method to convert a str or os.PathLike to the
+    proper Path or CloudPath object using AnyPath.
+    """
+    # shortcut pathlike items that are already valid Path/CloudPath
+    if isinstance(s, (CloudPath, Path)):
+        return s
+
+    return AnyPath(s)  # type: ignore
```
pjbull, line 66: @jayqi For some reason, this was the only way I could make `mypy` happy after the call here. Maybe you've got a nice solution:

https://github.com/drivendataorg/cloudpathlib/blob/8dca2edf4f950ee1452b4939e51a480f6b79b4a4/cloudpathlib/cloudpath.py#L695-L697

--------

If I change that line to `anypath.AnyPath(destination)`, I see these errors even if I `# type: ignore` that line.

```
mypy cloudpathlib
cloudpathlib/cloudpath.py:699: error: Item "str" of "Union[str, _PathLike[Any], CloudPath]" has no attribute "exists"
cloudpathlib/cloudpath.py:699: error: Item "_PathLike[Any]" of "Union[str, _PathLike[Any], CloudPath]" has no attribute "exists"
cloudpathlib/cloudpath.py:699: error: Item "str" of "Union[str, _PathLike[Any], CloudPath]" has no attribute "is_file"
cloudpathlib/cloudpath.py:699: error: Item "_PathLike[Any]" of "Union[str, _PathLike[Any], CloudPath]" has no attribute "is_file"
cloudpathlib/cloudpath.py:704: error: Item "str" of "Union[str, _PathLike[Any], CloudPath]" has no attribute "mkdir"
cloudpathlib/cloudpath.py:704: error: Item "_PathLike[Any]" of "Union[str, _PathLike[Any], CloudPath]" has no attribute "mkdir"
cloudpathlib/cloudpath.py:716: error: Incompatible return value type (got "Union[str, _PathLike[Any], CloudPath]", expected "Union[Path, CloudPath]")
Found 7 errors in 1 file (checked 24 source files)
```

pjbull:
@jayqi This is ready for re-review:
 - I opened issue for a few things that were too much to take on and block merging.
 - I'd like to have the default for upload _not_ overwrite a newer file on the cloud. Agreed that normal path functions don't typically care, but I do think overwriting newer versions on cloud storage warrants extra caution.
 - Changing some of the return types per the review was interwoven with a handful of other functions, and `mypy` was very fussy for a long time. I eventually got there, but had to add a little bit of a workaround to `anypath`. Would appreciate your eyes there.

PR #150, opened 2021-07-02
Support TransferConfig for s3 uploads and downloads

### LINKED ISSUE #147: Allow passing a TransferConfig to control upload/download configuration for S3
Boto3 supports a [`TransferConfig`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.TransferConfig) that can control multipart, threads, and more. We currently don't have any way to support setting this configuration.

I'd propose adding a `boto3_transfer_config` parameter to the `S3Client` init.  We can attach this config to the object and then pass it as `Config` any time we call the `download_file` or `upload_file` methods:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.download_file

We needed this when we were downloading a file with a size greater than the default `multipart_threshold` in the default `TransferConfig`. We were trying to write to a mounted disk that did not support append, so the download failed. We could have fixed this by setting `multipart_threshold` to greater than the size of the download, but we had no way to set this for `cloudpathlib`.

ejm714:
Adds the ability to specify [boto3.s3.transfer.TransferConfig](https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/s3/transfer.html#TransferConfig) parameters, which is useful for controlling multipart and thread use in uploads and downloads.

Outstanding:
- [x] test coverage

Closes #147

### DIFF
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -29,6 +29,7 @@ class S3Client(Client):
         boto3_session: Optional["Session"] = None,
         local_cache_dir: Optional[Union[str, os.PathLike]] = None,
         endpoint_url: Optional[str] = None,
+        boto3_transfer_config: Optional[dict] = None,
     ):
         """Class constructor. Sets up a boto3 [`Session`](
         https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html).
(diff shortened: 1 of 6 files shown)


pjbull:
@ejm714 Got a "live" test running with the code that you provided. Here's the diff if you are curious:
https://github.com/drivendataorg/cloudpathlib/pull/150/files/eee669a01883f61fd4199234cda183bb9a5fe22f..3942ac70e3d9e78d446efa0ac5d4cfbf13af3059

 - Bonus fix: give the test rigs a `live_server` flag so that we can branch within tests if necessary (like for this one).

I think this is ready to merge, thanks!

PR #156, opened 2021-08-20
Pipeline-ify the CI jobs

jayqi:
- Splits code quality check (linting) into a separate workflow job that runs before mocked tests. If code quality fails, mocked tests do not run.
- Merges the live tests into the main test workflow, but have it run after the mocked tests
- Add a workflow job that opens an issue if any previous job fails for non-PR workflows.


pjbull:
🎉 !

PR #164, opened 2021-09-07
Add no_sign_request for S3Client

### LINKED ISSUE #38: Support anonymous through boto to work with public S3 assets
Here's a first pass that could go in `__init__`.

The big downside is that it does a try-except with an actual network call. Ideally, there would be a way in `boto3` just to check if no creds were configured, and if not fall back to trying UNSIGNED.

```python
        # fallback to trying anonymous "UNSIGNED" requests, which can be used
        # if assets are public
        try:
            self.client.list_buckets()
        except NoCredentialsError:
            self.client = self.sess.client('s3', config=Config(signature_version=UNSIGNED))

        try:
            list(self.s3.buckets.limit(1))
        except NoCredentialsError:
            self.s3 = self.sess.resource("s3", config=Config(signature_version=UNSIGNED))
```

pjbull:
S3 requires explicitly not signing requests if users want anonymous access. To access a public bucket without any credentials configured, you need to instantiate a client that will not sign requests.

This PR adds a `no_sign_request` kwarg to `S3Client` that mimics the way a user might pass `--no-sign-request` with the AWS CLI to access public assets anonymously.

Closes #38

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,9 @@
 # cloudpathlib Changelog
 
+## v0.5.1 (unreleased)
+
+ - Fix #38 to allow passing `no_sign_request` to `S3Client` so that we can do anonymous requests for public resources on S3.
+
 ## v0.5.0 (2021-08-31)
 
 - Added `boto3_transfer_config` parameter to `S3Client` instantiation, which allows passing a `boto3.s3.transfer.TransferConfig` object and is useful for controlling multipart and thread use in uploads and downloads. See [documentation](https://cloudpathlib.drivendata.org/api-reference/s3client/#cloudpathlib.s3.s3client.S3Client.__init__) for more details. ([#150](https://github.com/drivendataorg/cloudpathlib/pull/150))
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,7 +2,7 @@
 
 ## v0.5.1 (unreleased)
 
- - Fix #38 to allow passing `no_sign_request` to `S3Client` so that we can do anonymous requests for public resources on S3.
+ - Added `no_sign_request` parameter to `S3Client` instantiation for anonymous requests for public resources on S3. See [documentation](https://cloudpathlib.drivendata.org/stable/api-reference/s3client/#cloudpathlib.s3.s3client.S3Client.__init__) for more details. ([#164](https://github.com/drivendataorg/cloudpathlib/pull/164))
 
 ## v0.5.0 (2021-08-31)
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,8 +1,8 @@
 # cloudpathlib Changelog
 
-## v0.5.1 (unreleased)
+## v0.6.0 (2021-09-07)
 
- - Added `no_sign_request` parameter to `S3Client` instantiation for anonymous requests for public resources on S3. See [documentation](https://cloudpathlib.drivendata.org/stable/api-reference/s3client/#cloudpathlib.s3.s3client.S3Client.__init__) for more details. ([#164](https://github.com/drivendataorg/cloudpathlib/pull/164))
+- Added `no_sign_request` parameter to `S3Client` instantiation for anonymous requests for public resources on S3. See [documentation](https://cloudpathlib.drivendata.org/stable/api-reference/s3client/#cloudpathlib.s3.s3client.S3Client.__init__) for more details. ([#164](https://github.com/drivendataorg/cloudpathlib/pull/164))
 
 ## v0.5.0 (2021-08-31)
 
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -62,9 +62,18 @@ class S3Client(Client):
             boto3_transfer_config (Optional[dict]): Instantiated TransferConfig for managing s3 transfers.
                 (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.TransferConfig)
         """
+        if boto3_session is not None:
+            self.sess = boto3_session
+        else:
+            self.sess = Session(
+                aws_access_key_id=aws_access_key_id,
+                aws_secret_access_key=aws_secret_access_key,
+                aws_session_token=aws_session_token,
+                botocore_session=botocore_session,
+                profile_name=profile_name,
+            )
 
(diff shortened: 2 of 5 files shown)


HISTORY.md
```diff
@@ -1,5 +1,9 @@
 # cloudpathlib Changelog
 
+## v0.5.1 (unreleased)
```
jayqi, line 3: I think we want this to be v0.6.0. Per Semver:

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> 1. MAJOR version when you make incompatible API changes,
> 2. **MINOR version when you add functionality in a backwards compatible manner**, and
> 3. PATCH version when you make backwards compatible bug fixes.

Also we probably just want to release after this is merged, so you can probably replace the `(unreleased)` with the date and bump the version in `setup.py`.

tests/test_s3_specific.py
```diff
@@ -130,3 +131,24 @@ def _execute_on_subprocess_and_observe(use_threads):
...
         assert _execute_on_subprocess_and_observe(use_threads=True) > 10
+
+
+def test_no_sign_request(s3_rig, tmp_path):
+    """Tests that boto3 receives and respects the transfer config
+    when working with a live backend. Does this by observing
+    if the `use_threads` parameter changes to number of threads
+    used by a child process that does a download.
+    """
+    if s3_rig.live_server:
```
jayqi, line 142: Maybe instead use
```
@pytest.mark.skipif(os.getenv("USE_LIVE_CLOUD") != "1")
```

Not a big deal. Maybe something we can decide to clean up later since there are other instances of this.

tests/test_s3_specific.py
```diff
@@ -130,3 +131,24 @@ def _execute_on_subprocess_and_observe(use_threads):
...
         # usually ~15 threads are spun up whe use_threads is True
         assert _execute_on_subprocess_and_observe(use_threads=True) > 10
+
+
+def test_no_sign_request(s3_rig, tmp_path):
+    """Tests that boto3 receives and respects the transfer config
+    when working with a live backend. Does this by observing
+    if the `use_threads` parameter changes to number of threads
+    used by a child process that does a download.
+    """
```
jayqi, line 141: Bad docstring copy-paste.

cloudpathlib/s3/s3client.py
```diff
@@ -57,18 +62,34 @@ def __init__(
...
         """
-        if boto3_session is not None:
-            self.sess = boto3_session
-        else:
-            self.sess = Session(
-                aws_access_key_id=aws_access_key_id,
-                aws_secret_access_key=aws_secret_access_key,
-                aws_session_token=aws_session_token,
-                botocore_session=botocore_session,
-                profile_name=profile_name,
```
jayqi, line 68: I think you actually want to keep the block of logic related to the session up here on the outside of `if no_sign_request`. 
- If `boto3_session` is passed, it should take precedence. 
- Otherwise it uses the arguments, which all default to `None`, which would be the same as what you have with `self.sess = Session()` in your `if no_sign_request` branch. 

```
if boto3_session is not None:
    # set passed session
else:
    # create session from args

if no_sign_request:
    # create s3 and client with UNSIGNED config
else:
    # create s3 and client without config
```

HISTORY.md
```diff
@@ -1,5 +1,9 @@
...
+
+ - Fix #38 to allow passing `no_sign_request` to `S3Client` so that we can do anonymous requests for public resources on S3.
```
jayqi, line 5:
```suggestion
 - Added `no_sign_request` parameter to `S3Client` instantiation for anonymous requests for public resources on S3. See [documentation](https://cloudpathlib.drivendata.org/stable/api-reference/s3client/#cloudpathlib.s3.s3client.S3Client.__init__) for more details. ([#164](https://github.com/drivendataorg/cloudpathlib/pull/164))
```

jayqi (changes requested):

tests/test_s3_specific.py
pjbull: For some tests I think we may want different checks if `rig.live_server` is True (e.g., we could have written this to confirm that config was passed to the Mock, but that didn't seem worth it), so I'm more inclined to use checking the rig to determine if we are live or not

cloudpathlib/s3/s3client.py
pjbull: Yeah that is how I started implementing it, but when looking at the AWS CLI docs they say the creds won't even be loaded if `--no-sign-request` is passed to the CLI:
https://awscli.amazonaws.com/v2/documentation/api/latest/reference/index.html

As implemented, we basically mimic that (ignore all creds and don't even load them if `no_sign_request`). Agreed that mixing and matching `no_sign_request` with any of the other options could potentially cause confusion.

There's not enough about UNSIGNED in the boto3 docs to track down how that config works with creds being passed as well so not sure what the implications of creating the session with user passed creds then not signing are.

pjbull: Both versions pass our test so will keep the session creds around if they are passed.

jayqi: I think the test should be enough, right? 

---

Went to look at what awscli uses, and they go through a slightly different mechanism in botocore I guess.

https://github.com/aws/aws-cli/blob/12956b24b88e0b60f6a871013d153f4e513f030f/awscli/customizations/globalargs.py#L87-L90

https://github.com/boto/botocore/blob/c75af12b09618a37b6c7f0bfac5f8cc6aa0bb7e7/botocore/handlers.py#L284-L289

pjbull: Yeah, we got this approach from their issues, so should be reliable:
https://github.com/boto/boto3/issues/1200

I tested both with and without a `~/.aws/credentials` file and confirmed that I could access public S3 items without any credentials when I passed the param

jayqi: In any case, I don't think your implementation actually adds any benefit in terms of not loading any credentials. If you initiate `Session()`, it just initializes with a default of `None` for everything, which will load from env vars.

pjbull: 🤷 maybe. I didn't dig deep enough in Boto3 to track when they do the actual loading and see if they pay attention to `UNSIGNED` when determining if they should even look for credentials.

jayqi (approved):

PR #165, opened 2021-09-07
Upload live test coverage to codecov

jayqi:
Right now only mocked tests are reflected in codecov.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,6 +113,12 @@ jobs:
           CUSTOM_S3_SECRET_KEY: ${{secrets.CUSTOM_S3_SECRET_KEY}}
           CUSTOM_S3_ENDPOINT: ${{secrets.CUSTOM_S3_ENDPOINT}}
 
+      - name: Upload coverage to codecov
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -113,6 +113,12 @@ jobs:
           CUSTOM_S3_SECRET_KEY: ${{secrets.CUSTOM_S3_SECRET_KEY}}
           CUSTOM_S3_ENDPOINT: ${{secrets.CUSTOM_S3_ENDPOINT}}
 
+      - name: Upload coverage to codecov
+        uses: codecov/codecov-action@v1
+        with:
+          file: ./coverage.xml
+          fail_ci_if_error: true
```
pjbull, line 120: Does codecov merge these for us rather than having two different reports?

pjbull: I guess if coverage goes up above then we've got a merge scenario

jayqi: Yes, codecov merges them.

jayqi: https://docs.codecov.com/docs/merging-reports

PR #166, opened 2021-09-08
Fix absolute URLs to point to stable/ docs

jayqi:
- Links in `README.md` and `HISTORY.md` are absolute to the `stable` docs, but get replaced with relative links when docs are built. 
- Replaced an absolute link in a docstring with a relative link. I'm a little iffier on this one, because now the raw docstring no longer has a useful valid link by itself.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,16 +1,20 @@
 # cloudpathlib Changelog
 
+## v0.6.1 (unreleased)
+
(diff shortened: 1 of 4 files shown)


jayqi:
Actually, I wonder if it's better to make these into relative links as much as possible...

pjbull (approved):

PR #168, opened 2021-09-17
Update _s3_file_query to check if an object exists explicitly. 

### LINKED ISSUE #169: Downloading public files with no_sign_request throws error
Our downloading code checks if an S3Path is a file or directory in order to know how to process the download. This calls `_s3_file_query` which checks for existing using `Bucket.objects.filter`.

This call requires list permissions to the bucket, so we need another way to check if an S3Path is a file to support downloading assets without S3 credentials configured.

Here's an example traceback:
```
----> 1 download_weights(filename="zamba_time_distributed.ckpt", destination_dir=Path.cwd())

~/zamba-algorithms/zamba_algorithms/models/utils.py in download_weights(filename, weight_region, destination_dir)
     38         client=S3Client(local_cache_dir=destination_dir, no_sign_request=True),
     39     )
---> 40     return s3p.fspath

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/cloudpathlib/cloudpath.py in fspath(self)
    304     @property
    305     def fspath(self) -> str:
--> 306         return self.__fspath__()
    307 
    308     def glob(self, pattern: str) -> Iterable["CloudPath"]:

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/cloudpathlib/cloudpath.py in __fspath__(self)
    208 
    209     def __fspath__(self):
--> 210         if self.is_file():
    211             self._refresh_cache(force_overwrite_from_cloud=False)
    212         return str(self._local)

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/cloudpathlib/s3/s3path.py in is_file(self)
     37 
     38     def is_file(self) -> bool:
---> 39         return self.client._is_file_or_dir(self) == "file"
     40 
     41     def mkdir(self, parents=False, exist_ok=False):

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/cloudpathlib/s3/s3client.py in _is_file_or_dir(self, cloud_path)
    117 
    118         # get first item by listing at least one key
--> 119         s3_obj = self._s3_file_query(cloud_path)
    120 
    121         if s3_obj is None:

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/cloudpathlib/s3/s3client.py in _s3_file_query(self, cloud_path)
    131     def _s3_file_query(self, cloud_path: S3Path):
    132         """Boto3 query used for quick checks of existence and if path is file/dir"""
--> 133         return next(
    134             (
    135                 obj

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/cloudpathlib/s3/s3client.py in <genexpr>(.0)
    132         """Boto3 query used for quick checks of existence and if path is file/dir"""
    133         return next(
--> 134             (
    135                 obj
    136                 for obj in (

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/boto3/resources/collection.py in __iter__(self)
     81 
     82         count = 0
---> 83         for page in self.pages():
     84             for item in page:
     85                 yield item

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/boto3/resources/collection.py in pages(self)
    164         # we start processing and yielding individual items.
    165         count = 0
--> 166         for page in pages:
    167             page_items = []
    168             for item in self._handler(self._parent, params, page):

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/botocore/paginate.py in __iter__(self)
    253         self._inject_starting_params(current_kwargs)
    254         while True:
--> 255             response = self._make_request(current_kwargs)
    256             parsed = self._extract_parsed_response(response)
    257             if first_request:

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/botocore/paginate.py in _make_request(self, current_kwargs)
    330 
    331     def _make_request(self, current_kwargs):
--> 332         return self._method(**current_kwargs)
    333 
    334     def _extract_parsed_response(self, response):

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/botocore/client.py in _api_call(self, *args, **kwargs)
    384                     "%s() only accepts keyword arguments." % py_operation_name)
    385             # The "self" in this scope is referring to the BaseClient.
--> 386             return self._make_api_call(operation_name, kwargs)
    387 
    388         _api_call.__name__ = str(py_operation_name)

~/anaconda3/envs/zamba-algorithms/lib/python3.8/site-packages/botocore/client.py in _make_api_call(self, operation_name, api_params)
    703             error_code = parsed_response.get("Error", {}).get("Code")
    704             error_class = self.exceptions.from_code(error_code)
--> 705             raise error_class(parsed_response, operation_name)
    706         else:
    707             return parsed_response

ClientError: An error occurred (AccessDenied) when calling the ListObjects operation: Access Denied
```

pjbull:
In  our current setup, the `_is_file_or_dir` check relies on `_s3_file_query`, which requires list permissions. When we pass `no_sign_request`, we want to be able to download public files without credentials. However, we're not able to use this functionality because we do not have list permissions to the bucket.

This change introduces a first check to see if an object exists using `load` (which just calls `head_object` so should have very little overhead and does not require list permissions). If that does not work, we do our normal call to see if it is a dir.

Downside:
1 extra network call in the case of a directory.

Bonuses:
 - Update `no_sign_request` test to include scenario that failed before this fix.
 - We used `get` which downloads the file where we should have used `load`. Fixed this instance.
 - Some other tests were failing because our test bucket had not been cleaned up. I changed these tests to properly use the `rig.test_dir` folder so state on the bucket shouldn't exist. Also removed cleanup performed by `_download_with_threads` in the test since it should be handled by the rig now.

Closes #169

### DIFF
--- a/tests/mock_clients/mock_s3.py
+++ b/tests/mock_clients/mock_s3.py
@@ -5,6 +5,7 @@ import shutil
 from tempfile import TemporaryDirectory
 
 from boto3.session import Session
+from botocore.exceptions import ClientError
 
 from .utils import delete_empty_parents_up_to_root
 
@@ -65,6 +66,16 @@ class MockBoto3Object:
             raise NoSuchKey({}, {})
         else:
             return {"key": self.path}
+
+    def load(self):
+        if not self.path.exists() or self.path.is_dir():
+            raise ClientError({}, {})
+        else:
+            return {"key": str(self.path)}
+
+    @property
+    def key(self):
+        return str(self.path.relative_to(self.root))
 
     def copy_from(self, CopySource=None, Metadata=None, MetadataDirective=None):
         if CopySource["Key"] == str(self.path.relative_to(self.root)):
--- a/tests/mock_clients/mock_s3.py
+++ b/tests/mock_clients/mock_s3.py
@@ -65,17 +65,17 @@ class MockBoto3Object:
         if not self.path.exists() or self.path.is_dir():
             raise NoSuchKey({}, {})
         else:
-            return {"key": self.path}
+            return {"key": str(PurePosixPath(self.path))}
 
     def load(self):
         if not self.path.exists() or self.path.is_dir():
             raise ClientError({}, {})
         else:
-            return {"key": str(self.path)}
+            return {"key": str(PurePosixPath(self.path))}
 
     @property
     def key(self):
-        return str(self.path.relative_to(self.root))
+        return str(PurePosixPath(self.path).relative_to(PurePosixPath(self.root)))
 
     def copy_from(self, CopySource=None, Metadata=None, MetadataDirective=None):
         if CopySource["Key"] == str(self.path.relative_to(self.root)):
--- a/tests/test_s3_specific.py
+++ b/tests/test_s3_specific.py
@@ -56,37 +56,31 @@ def _download_with_threads(s3_rig, tmp_path, use_threads):
     """Job used by tests to ensure Transfer config changes are
     actually passed through to boto3 and respected.
     """
-    try:
-        sleep(1)  # give test monitoring process time to start watching
+    sleep(1)  # give test monitoring process time to start watching
 
-        transfer_config = TransferConfig(
-            max_concurrency=100,
-            use_threads=use_threads,
-            multipart_chunksize=1 * 1024,
-            multipart_threshold=10 * 1024,
-        )
-        client = s3_rig.client_class(boto3_transfer_config=transfer_config)
-        p = client.CloudPath(f"s3://{s3_rig.drive}/{s3_rig.test_dir}/dir_0/file0_to_download.txt")
+    transfer_config = TransferConfig(
+        max_concurrency=100,
+        use_threads=use_threads,
+        multipart_chunksize=1 * 1024,
+        multipart_threshold=10 * 1024,
+    )
+    client = s3_rig.client_class(boto3_transfer_config=transfer_config)
+    p = client.CloudPath(f"s3://{s3_rig.drive}/{s3_rig.test_dir}/dir_0/file0_to_download.txt")
 
-        assert not p.exists()
+    assert not p.exists()
 
-        # file should be about 60KB
-        text = "lalala" * 10_000
-        p.write_text(text)
+    # file should be about 60KB
+    text = "lalala" * 10_000
+    p.write_text(text)
 
-        assert p.exists()
+    assert p.exists()
 
-        # assert not (dl_dir / p.name).exists()
-        p.download_to(tmp_path)
+    # assert not (dl_dir / p.name).exists()
+    p.download_to(tmp_path)
 
-        p.unlink()
+    p.unlink()
 
-        assert not p.exists()
-
-    finally:
-        p = s3_rig.create_cloud_path("dir_0/file0_0.txt")
-        if p.exists():
-            p.unlink()
+    assert not p.exists()
 
 
 def test_transfer_config_live(s3_rig, tmp_path):
--- a/tests/test_s3_specific.py
+++ b/tests/test_s3_specific.py
@@ -1,5 +1,7 @@
 from concurrent.futures import ProcessPoolExecutor
 from time import sleep
+from pathlib import Path
+
 import pytest
 
 from boto3.s3.transfer import TransferConfig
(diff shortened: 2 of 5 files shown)


tests/mock_clients/mock_s3.py
```diff
@@ -66,6 +67,16 @@ def get(self):
...
 
+    def load(self):
+        if not self.path.exists() or self.path.is_dir():
+            raise ClientError({}, {})
+        else:
+            return {"key": str(self.path)}
+
+    @property
+    def key(self):
+        return str(self.path.relative_to(self.root))
```
jayqi: Tests might be failing because you're calling `str` here. I suspect it's a `PosixPath` vs `WindowsPath` thing.

tests/test_s3_specific.py
```diff
@@ -29,7 +31,9 @@ def test_transfer_config(s3_rig, tmp_path):
 
     # download
     client.set_as_default_client()
-    p = s3_rig.create_cloud_path("dir_0/file0_0.txt")
+    p = s3_rig.create_cloud_path(f"{s3_rig.test_dir}/dir_0/file0_0.txt")
```
jayqi, line 34: I don't think you want to add these here. `s3_rig.create_cloud_path` should already include the `test_dir`.

https://github.com/drivendataorg/cloudpathlib/blob/ea3053f24f6425b12662df5489ee5f1f57452875/tests/conftest.py#L77-L82

tests/test_s3_specific.py
```diff
@@ -39,7 +43,7 @@ def test_transfer_config(s3_rig, tmp_path):
         assert client.s3.download_config == transfer_config
 
     # upload
-    p2 = s3_rig.create_cloud_path("dir_0/file0_0_uploaded.txt")
+    p2 = s3_rig.create_cloud_path(f"{s3_rig.test_dir}/dir_0/file0_0_uploaded.txt")
```
jayqi, line 46: Same as previous comment that `s3_rig.create_cloud_path` should already handle `test_dir`.

tests/test_s3_specific.py
```diff
@@ -29,7 +31,9 @@ def test_transfer_config(s3_rig, tmp_path):
 
     # download
     client.set_as_default_client()
-    p = s3_rig.create_cloud_path("dir_0/file0_0.txt")
+    p = s3_rig.create_cloud_path(f"{s3_rig.test_dir}/dir_0/file0_0.txt")
+    p.write_text("test file")
```
jayqi, line 35: If you keep the original `s3_rig.create_cloud_path("dir_0/file0_0.txt")` then `dir_0/file0_0.txt` should already exist because it's copied over when the rig is set up.

tests/test_s3_specific.py
```diff
@@ -133,10 +137,11 @@ def _execute_on_subprocess_and_observe(use_threads):
         assert _execute_on_subprocess_and_observe(use_threads=True) > 10
 
 
-def test_no_sign_request(s3_rig):
+def test_no_sign_request(s3_rig, tmp_path):
     """Tests that we can pass no_sign_request to the S3Client and we will
     be able to access public resources but not private ones.
     """
+    tmp_path = Path(tmp_path)
```
jayqi, line 144: `tmp_path` is already a `Path` object. ([docs](https://docs.pytest.org/en/stable/tmpdir.html#the-tmp-path-fixture))

tests/test_s3_specific.py
```diff
@@ -82,7 +86,7 @@ def _download_with_threads(s3_rig, tmp_path, use_threads):
         assert not p.exists()
 
     finally:
-        p = s3_rig.create_cloud_path("/dir_0/file0_0.txt")
+        p = s3_rig.create_cloud_path(f"{s3_rig.test_dir}/dir_0/file0_0.txt")
```
jayqi, line 89: I'm not sure what the purpose of this `finally` block is for this test, but again, same comment that `s3_rig.create_cloud_path` already handles `test_dir`.

jayqi (changes requested):

tests/test_s3_specific.py
pjbull: Looks like cleanup since in order to pass the TransferConfig we have a file created outside of the test rig:
https://github.com/drivendataorg/cloudpathlib/blob/7f6943265ae6fc98a2595b28cb98d996672fd209/tests/test_s3_specific.py#L70-L71

tests/mock_clients/mock_s3.py
pjbull: Fixed 👍

tests/test_s3_specific.py
jayqi: That's not even the same filename though. `file0_0.txt` vs. `file0_to_download.txt`.

pjbull: 🤔 good point. I'll try pulling it out, but it's not really part of this change so will file an issue if something breaks

pjbull: Ran twice locally and watched the actual S3 bucket to confirm it was cleaned up after the test ran. I think should be good to go now.

pjbull:
Thanks @jayqi should be good now!

jayqi (approved):

PR #177, opened 2021-10-06
Fix broken build mypy errors

### LINKED ISSUE #172: Failed build on master branch (tests #417)
Workflow failed: [tests #417](https://github.com/drivendataorg/cloudpathlib/actions/runs/1274005771)

jayqi:
1. Updates `AzureBlobClient._get_metadata` return type per discussion in #172 
2. Explicitly silences mypy errors from `azure.storage.blob.*`. I decided we should implement this workaround because neither of these two following things seem like they're on track to getting fixed. 
    - The Azure SDK typing errors are out of our control, but the upstream issue hasn't had any news in 9 days. https://github.com/Azure/azure-sdk-for-python/issues/20771
    - There is some kind of bug in mypy, as site-packages errors are supposed to be ignored but aren't. https://github.com/python/mypy/issues/11239#issuecomment-934479522

Closes #172

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,6 +14,11 @@ exclude =
 ignore_missing_imports = True
 allow_redefinition = True
 
+# Temporarily silence Azure SDK bug https://github.com/Azure/azure-sdk-for-python/issues/20771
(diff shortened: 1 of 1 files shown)


pjbull:
Thanks @jayqi!

PR #178, opened 2021-10-20
Fix small bug in mkdocs

klwetstone:
- Fixes incorrect reference in `watch` of `mkdocs.yml` so building docs locally works
- Adds a makefile command for `make docs-serve` that serves the docs on a site locally

The standard mkdocs project setup actually has `mkdocs.yml` in the project root, and would build `docs/` into `site/` (as opposed to `docs/docs/` into `docs/site/`). We've wrapped things into a subdirectory layer of `docs/` that is non-standard to keep the root level cleaner (credit to Jay for the explanation!). To refer to the right place when building docs locally, we need:
```
watch:
    ../cloudpathlib
```
Instead of
```
watch:
   cloudpathlib
```
In `mkdocs.yml`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,20 +38,17 @@ dist: clean ## builds source and wheel package
 	python setup.py bdist_wheel
 	ls -l dist
 
-docs: clean-docs
+docs-setup:  ## setup docs pages based on README.md and HISTORY.md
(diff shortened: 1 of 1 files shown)


klwetstone:
@jayqi this fixes the same issue in cloudpathlib that we just fixed in zamba if you have a chance to look it over. Thanks!

klwetstone:
@jayqi updated! should be good to merge

PR #185, opened 2021-12-29
Fix nested remove for azure and gcs

### LINKED ISSUE #184: rmtree not working when there is directory inside (google cloud storage)
This package is really helpful to work with different cloud providers but I'm facing an issue when trying to remove directory in gcs.

I tried it even in some test script using this line (after handle the authentication, because I do get True for methods like .exists() or is_dir().

```CloudPath("gs://<bucket>/<path_to_folder(prefix)>").rmtree()```

and getting exception from the  _remove function. ('NoneType' object has no attribute 'delete').

Am I missing something in my usage or maybe it is knows bug?
Thanks

pjbull:
Closes #184 

Problem is that `_list_dir` implementations return both directories and files, but our `_remove` function for both of these clients assumes only files. 

Added regression test which fails on Azure + GCS without this fixes, but passes with the fix.

Updated history and version to cut new release with this fix

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,11 @@
 # cloudpathlib Changelog
 
+## v0.6.3 (2021-12-29)
+
+- Fixed error when using `rmtree` on nested directories for Google Cloud Storage and Azure Blob Storage. ([Issue #184](https://github.com/drivendataorg/cloudpathlib/issues/184), [PR #185](https://github.com/drivendataorg/cloudpathlib/pull/185))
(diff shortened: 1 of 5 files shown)


pjbull:
@jayqi Going to merge this since I think this is low-risk and well-tested so that we can cut a release to fix #184.

Happy to address any code review comments in separate PR.

jayqi:
Looks okay. The main thing that stands out to me is that this may have a performance impact from additional network calls.

PR #193, opened 2022-01-25
Add ability to set endpoint_url from AWS_ENDPOINT_URL env variable

kabirkhan:
Hey guys! Long time :)

Loving this package so far. One minor convenience that might be easier to merge here than this ridiculous upstream PR for the boto3 library.

Allows setting the endpoint_url from the AWS_ENDPOINT_URL env variable while waiting for the upstream PR from boto3: https://github.com/boto/boto3/pull/2746

I'll add a test if the idea works for you guys

### DIFF
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -65,7 +65,6 @@ class S3Client(Client):
                 (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.TransferConfig)
         """
         endpoint_url = endpoint_url or os.getenv("AWS_ENDPOINT_URL")
-        print("ENDPOINT URL IN INIT:", endpoint_url)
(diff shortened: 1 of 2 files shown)


pjbull:
Looks good to me! Test would be great, and then an issue to track removing if/when the upstream issue is resolved.

kabirkhan:
Here's that test

kabirkhan:
Added the issue as well

pjbull:
Test looks good. Can you make sure linting passes? (`make lint` should do that for you locally)

kabirkhan:
Done, not sure why the live tests would be failing though. Doesn't seem to be related to this change.

pjbull:
@kabirkhan yeah, i think that is just that PRs from forks can't use the secrets so live tests can't authenticate. i'll do some digging to see how we enable that.

pjbull:
Dropped these changes on their own branch so we can run the live tests there to confirm

PR #199, opened 2022-02-03
Fix S3 recursion for fake directory

### LINKED ISSUE #198: Recursion error on "fake" s3 directories
This was brought up by @remi-braun in #148.

Sometimes, listing a directory with the `S3Client` results in a recursion error. This was seen on Ceph where "faked" S3 directories are more common.

@GeorgeSabu found that this is likely the same root cause as what was fixed in #190. We assume that all the "content" items from the API are files, but they can be directories as well. The way to detect which is a directory is to see if the size is 0, as is implemented here:
https://github.com/ElucidataInc/cloudpathlib/commit/9c11c6af2b0ac713ddcd950123d1db3b17515efa

GeorgeSabu:
Closes #198

### DIFF
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -197,7 +197,7 @@ class S3Client(Client):
 
                 # files in the directory
                 for result_key in result.get("Contents", []):
-                    if result_key.get('Size') > 0:
+                    if result_key.get("Size") > 0:
                         yield self.CloudPath(f"s3://{cloud_path.bucket}/{result_key.get('Key')}")
 
     def _move_file(self, src: S3Path, dst: S3Path, remove_src: bool = True) -> S3Path:
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -197,7 +197,8 @@ class S3Client(Client):
 
                 # files in the directory
                 for result_key in result.get("Contents", []):
-                    yield self.CloudPath(f"s3://{cloud_path.bucket}/{result_key.get('Key')}")
+                    if result_key.get('Size') > 0:
+                        yield self.CloudPath(f"s3://{cloud_path.bucket}/{result_key.get('Key')}")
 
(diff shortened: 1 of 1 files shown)


pjbull:
@GeorgeSabu Can you update the tests to pass? Looks like you just need to run `black` to format the file you changed (you can use the `make lint` command too).

pjbull:
@GeorgeSabu Yep, that'd be great! You can re-open this PR or start a new one. (Sorry, I wasn't paying attention and clicked thought I was clicking the button the run the tests, but hit the merge one instead.)

I thought it was going to be a simple fix to the tests, but it seems a little more involved. At the very least, you'll want to update the `MockBoto3Paginator` here to work like the real S3 API results look like. Here's where it is defined:
https://github.com/drivendataorg/cloudpathlib/blob/f951e21e5480065a93eb03e8459ffea2f40f4921/tests/mock_clients/mock_s3.py/#L192-L226

And in a quick look, I think you'll at least want something like this that adds the `Size` key and creates a "fake" directory like we see on S3.

```python
class MockBoto3Paginator:
    def __init__(self, root, per_page=2):
        self.root = root
        self.per_page = per_page

    def paginate(self, Bucket=None, Prefix="", Delimiter=None):
        new_dir = self.root / Prefix

        items = [f for f in new_dir.iterdir() if not f.name.startswith(".")]

        for ix in range(0, len(items), self.per_page):
            page = items[ix : ix + self.per_page]
            dirs = [
                {"Prefix": str(_.relative_to(self.root).as_posix())} for _ in page if _.is_dir()
            ]
            files = [
                {
                    "Key": str(_.relative_to(self.root).as_posix()),
                    "Size": 123
                    if not _.relative_to(self.root).exists()
                    else _.relative_to(self.root).stat().st_size,
                }
                for _ in page
                if _.is_file()
            ]

            # s3 can have "fake" directories where size is 0, but it is listed in "Contents" (see #198)
            # add one in here for testing
            if dirs:
                fake_dir = dirs.pop(-1)
                fake_dir["Size"] = 0
                files.append(fake_dir)

            yield {"CommonPrefixes": dirs, "Contents": files}

```

Issue #181, opened 2021-12-08
Redundant unreachable code that cause high error rate in S3.


fadikhou:
Hello, 

Recently I started using cloudpath library to upload entire directories to S3, but suddenly I noticed a weird behavior that we got a high error rate at S3 (according to AWS monitors) with the following errors:
botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.
botocore.exceptions.ClientError: An error occurred (404) when calling the HeadObject operation: Not Found

And after a deep investigation I have found that those errors came from cloudpath.py::upload_from method since the library checks for each file whether it is exists or not before uploading it (**if self.exists() and self.is_dir()**)

    def upload_from(
        self, source: Union[str, os.PathLike], force_overwrite_to_cloud: bool = False
    ) -> "CloudPath":
        """Upload a file or directory to the cloud path."""
        source = Path(source)

        if source.is_dir():
            for p in source.iterdir():
                (self / p.name).upload_from(p, force_overwrite_to_cloud=force_overwrite_to_cloud)

            return self

        else:
            **if self.exists() and self.is_dir():
                dst = self / source.name**
            else:
                dst = self

            dst._upload_file_to_cloud(source, force_overwrite_to_cloud=force_overwrite_to_cloud)

            return dst

My question is: 
Why we need this check ? its redundant because self.is_dir() always returns False.

jayqi:
> Why we need this check ? its redundant because self.is_dir() always returns False.

Am I understanding you correctly that you think the `self.is_dir()` isn't necessary when `self.exists()` is `False`? I think this specifically isn't actually a problem. Python's `and` operator will short-circuit, which means if `self.exists()` evaluates to `False`, it doesn't actually even execute `self.is_dir()`. Documentation [here](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not), 

The check for `self.is_dir()` is needed in general (if `self.exists()` is `True`) to match the expected behavior when you try to move a file to a directory. If your source is a file and destination is a directory, most programs will put your file _into_ that directory, instead of moving your file to where that directory is. 

---

> high error rate at S3 (according to AWS monitors) with the following errors:
> botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.
> botocore.exceptions.ClientError: An error occurred (404) when calling the HeadObject operation: Not Found

[...]

Issue #176, opened 2021-10-06
Speed up `is_dir` on entries returned by `iterdir` and `glob`


analog-cbarber:
The `is_dir` check is fairly expensive, but at least for S3 and Azure when the entries were created as a result of the client's `_list_dir` method, you can tell for each entry whether it is a directory or a file and immediately set the result on the created CloudPath instance.

For example for the `S3Client._list_dir`, you could write something like:

```python
            paginator = self.client.get_paginator("list_objects_v2")

            for result in paginator.paginate(
                Bucket=cloud_path.bucket, Prefix=prefix, Delimiter="/", MaxKeys=1000
            ):

                # sub directory names
                for result_prefix in result.get("CommonPrefixes", []):
                    path = S3Path(f"s3://{cloud_path.bucket}/{result_prefix.get('Prefix')}")
                    path._is_dir = True
                    yield path

                # files in the directory
                for result_key in result.get("Contents", []):
                    path = S3Path(f"s3://{cloud_path.bucket}/{result_key.get('Key')}")
                    path._is_dir = False
                    yield path
```

and modify `S3Path.is_dir`:

```python
    def is_dir(self) -> bool:
        if self._is_dir is None:
            self._is_dir = self.client._is_file_or_dir(self) == "dir"
        return self._is_dir
```

This makes a HUGE performance difference if you need to call `is_dir` on the entries returned from `iterdir` or `glob` (in my case, when implementing a file dialog that works for cloud paths).

Not sure if this particular implementation is the best way to do this, but something like this is needed.

analog-cbarber:
BTW, it probably would also be a good idea to cache other information returned in the directory entries that could be used to populate the `stat()` call, e.g. the file size and last modified time and perhaps the etag as well.

analog-cbarber:
Note that methods that create an entry should make sure to fix the cached state appropriately.

PR #202, opened 2022-02-05
Re-implement glob logic

### LINKED ISSUE #15: Confirm that all backends properly iterate over directories
Most cloud services don't have a notion of a "directory." We try to add this back in within our logic by doing our own path splitting.

We should make sure that the pathlib methods (`glob` and `iterdir`) and our methods do the same thing of:
 - Returning directories and/or files when iterating
 - Returning the root or current directory when being asked to iter or not
 - Sort order of the returned results (may not be possible/worth it since we use a generator)

### LINKED ISSUE #154: glob: difference between CloudPath and Paths
Hello,

I have a replicated filetree between my S3 compatible storage and a local filesystem.
However, there is a discrepency when I use the glob function.
The filesystem glob works fine, but I have to add `"**/*/"` at the beginning of the path for CloudPath's glob

```python
# Composition of the bucket (sorry for the long names)
# NB: there are duplicates with and without /, may be related to #148
>> list(cloud_path.glob("**/*"))
[S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/'), 
S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/DAX/'), 
S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/DAX'), 
S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/DAX/BY_PRODUCTS/'), 
S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/DAX/BY_PRODUCTS'), 
S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/DAX/BY_PRODUCTS/20191215T110441_S2_T30TXP_L2A_122756/'), 
S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/DAX/BY_PRODUCTS/20191215T110441_S2_T30TXP_L2A_122756'), 
S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/DAX/BY_PRODUCTS/20191215T110441_S2_T30TXP_L2A_122756/14_20191215T110441_S2_T30TXP_L2A_122756_CLIP.tif')]

# Not working ways of retrieving the file 14_20191215T110441_S2_T30TXP_L2A_122756_CLIP.tif
>> list(cloud_path.glob("DAX/BY_PRODUCTS/*/14_20191215T110441_S2_T30TXP_L2A_122756_CLIP.tif"))
[]
/14_20191215T110441_S2_T30TXP_L2A_122756_CLIP.tif')]
>> list(cloud_path.glob("**/DAX/BY_PRODUCTS/*/14_20191215T110441_S2_T30TXP_L2A_122756_CLIP.tif"))
[]
>> list(cloud_path.glob("*DAX/BY_PRODUCTS/*/14_20191215T110441_S2_T30TXP_L2A_122756_CLIP.tif"))
[]

# Correct way
>> list(cloud_path.glob("**/*DAX/BY_PRODUCTS/*/14_20191215T110441_S2_T30TXP_L2A_122756_CLIP.tif"))
[S3Path('s3://my_bucket/MODULES/PROCESS/CLIPPING/DAX/BY_PRODUCTS/20191215T110441_S2_T30TXP_L2A_122756]
```

Note: I am still on my Ceph compatible storage (see #148), I hope this is not related 😓

### LINKED ISSUE #186: Drop Python 3.6 support after 2021-12
Python 3.6 will stop receiving security updates after 2021-12:
https://www.python.org/dev/peps/pep-0494/#lifespan

We can remove it from our test matrix and our list of supported python versions in setup.py

We should replace it with Python 3.10 instead

pjbull:
`glob`, `rglob`, and `iterdir` did not work as their pathlib counterparts did.

In order to align our implementation with the `pathlib` versions, we adapted `CloudPaths` to work in the same way that the CPython implementation of `glob` does, [which is relatively complex in order to handle all of the glob scenarios](https://github.com/python/cpython/blob/3.10/Lib/pathlib.py#L385-L500). This is based on the `_Selector` class.

This change introduces two internal classes, `_CloudPathSelectable` and `_CloudPathSelectableAccessor` that wrap a `CloudPath` to expose the methods expected by `_Selector` and its children.

The change also changes the internal `Client._list_dir` API to return a tuple of the form (entry, is_dir), which could help address #181 and #176. Internal code where we use `iterdir` can be replaced with `_list_dir` if knowing if something is a file or directory is done within the method.

In order to close #15, this also makes sure that `iterdir` does not include `self`.

In order to make the tests pass, this also closes #186, which drops support for Python 3.6

Tests for all of these changes are included. The test suite is created to mirror the tests for `glob` and `rglob` in the CPython codebase. Note: `test_glob_many_open_files`, which emulates a CPython tests is pretty slow. We may consider not matching to the same depth and range to speed that up if it becomes frustrating.

Note: This does leverage some internals of CPython to work correctly. It's possible that implementations without the `_make_selector, _posix_flavour` private apis will not work. If this is a problem, we may need to re-implement or vendor `_Selector` and its child classes. It seems to at least work fine on `pypy` not sure about anything else:

```python
cloudpathlib on  glob-new-implementation [$?] 🐍 v3.7.12 via 🅒 cloudpathlib-pypy took 2s
❯ pypy
Python 3.7.12 | packaged by conda-forge | (44db2626, Nov 15 2021, 13:00:09)
[PyPy 7.3.7 with GCC Clang 11.1.0] on darwin

>>>> from cloudpathlib import CloudPath
>>>> root_dir = CloudPath("s3://drivendata-public-assets/")
>>>> list(root_dir.glob('**/*.txt'))
[S3Path('s3://drivendata-public-assets/nested_dir/test.dot/a_file.txt'), S3Path('s3://drivendata-public-assets/odsc-west-2019/DATA_DICTIONARY.txt')]
```

Closes #154
Closes #15
Closes #186


cloudpathlib/cloudpath.py
```diff
@@ -305,28 +305,57 @@ def exists(self) -> bool:
...
-    def glob(self, pattern: str) -> Iterable["CloudPath"]:
-        # strip cloud prefix from pattern if it is included
-        if pattern.startswith(self.cloud_prefix):
-            pattern = pattern[len(self.cloud_prefix) :]
+    def _glob_checks(self, pattern):
+        if self.is_file():
+            raise NotImplementedError("Cannot run glob with a file as the root.")
+
+        if ".." in pattern:
+            raise NotImplementedError("Relative paths with '..' not supported in glob patterns.")
```
jayqi, line 313: I think we should make a new class in [`cloudpathlib.exceptions`](https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/exceptions.py) like

```python
class CloudPathNotImplementedError(CloudPathException, NotImplementedError):
    pass
```

and use that.

cloudpathlib/cloudpath.py
```diff
@@ -305,28 +305,57 @@ def exists(self) -> bool:
...
 
-        # identify if pattern is recursive or not
-        recursive = False
-        if pattern.startswith("**/"):
-            pattern = pattern.split("/", 1)[-1]
-            recursive = True
+    def glob(self, pattern):
+        """Iterate over this subtree and yield all existing files (of any
+        kind, including directories) matching the given relative pattern.
+        """
```
jayqi, line 335: We're copying docstrings from pathlib so I don't think we need this.

cloudpathlib/cloudpath.py
```diff
@@ -305,28 +305,57 @@ def exists(self) -> bool:
...
+        pattern_parts = PurePosixPath(pattern).parts
+        selector = _make_selector(tuple(pattern_parts), _posix_flavour)
+
+        yield from self._glob(selector)
+
+    def rglob(self, pattern):
+        """Recursively yield all existing files (of any kind, including
+        directories) matching the given relative pattern, anywhere in
+        this subtree.
+        """
```
jayqi, line 347: Ditto on not needing the docstring.

cloudpathlib/cloudpath.py
```diff
@@ -866,3 +892,81 @@ def _resolve(path: PurePosixPath) -> str:
...
+# Designed to be compatible when used by these selector implementations from pathlib:
+# https://github.com/python/cpython/blob/3.10/Lib/pathlib.py#L385-L500
+class _CloudPathSelectableAccessor:
+    def __init__(self, scandir_func):
+        self.scandir = scandir_func
+
+
+class _CloudPathSelectable:
+    def __init__(
+        self, relative_cloud_path: PurePosixPath, children: List, is_dir: bool, exists: bool = True
```
jayqi, line 913:
```suggestion
        self, relative_cloud_path: PurePosixPath, children: Dict[PurePosixPath, bool], is_dir: bool, exists: bool
```

- Fix type hint for `children`
- I think we don't need the default on `exists`. I think we always specify, and that's probably safer anyways.

cloudpathlib/cloudpath.py
```diff
@@ -866,3 +892,81 @@ def _resolve(path: PurePosixPath) -> str:
...
+        return self._path.name
+
+    @staticmethod
+    @contextmanager
+    def scandir(root):
+        results = []
+        for c, is_dir in root._all_children.items():
+            if c.parent == root._path:
+                results.append(root._make_child_relpath(c.name))
+        yield results
```
jayqi, line 946:
```suggestion
        yield (
            root._make_child_relpath(c.name)
            for c, _ in root._all_children.items()
            if c.parent == root._path
        )
```

I think this should also work? Not sure if having a generator expression is clearer than the for loop, or if turning the list into an iterator is a meaningful optimization.

requirements-dev.txt
```diff
@@ -16,7 +16,7 @@ pandas
 pillow
 psutil
 pydantic
-pytest
+pytest<7.0
```
jayqi, line 19: What happened here? Were there breaking changes?

setup.py
```diff
@@ -59,5 +60,5 @@ def load_requirements(path: Path):
         "Source Code": "https://github.com/drivendataorg/cloudpathlib",
     },
     url="https://github.com/drivendataorg/cloudpathlib",
-    version="0.6.5",
+    version="0.6.6",
```
jayqi, line 63: Should match release notes.

jayqi (changes requested):

cloudpathlib/cloudpath.py
```diff
@@ -305,28 +305,57 @@ def exists(self) -> bool:
     def fspath(self) -> str:
         return self.__fspath__()
 
-    def glob(self, pattern: str) -> Iterable["CloudPath"]:
-        # strip cloud prefix from pattern if it is included
-        if pattern.startswith(self.cloud_prefix):
-            pattern = pattern[len(self.cloud_prefix) :]
+    def _glob_checks(self, pattern):
+        if self.is_file():
+            raise NotImplementedError("Cannot run glob with a file as the root.")
```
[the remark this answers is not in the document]
pjbull: Agreed. Removed this check and the tests. Confirmed you can both read and glob `foo`:

image

requirements-dev.txt
pjbull: Yeah, they just released 7.0 a couple days ago and our test suite breaks. There wasn't a simple/obvious fix to the error I was seeing. I'll file an issue to investigate.

pjbull:
@jayqi review comments implemented, thanks!

jayqi (approved):
I'm still pretty amazed that we got this to work. Exciting stuff!

cloudpathlib/s3/s3client.py
```diff
@@ -179,25 +179,30 @@ def _list_dir(self, cloud_path: S3Path, recursive=False) -> Iterable[S3Path]:
...
+                        True,
+                    )
 
                 # files in the directory
                 for result_key in result.get("Contents", []):
-                    yield self.CloudPath(f"s3://{cloud_path.bucket}/{result_key.get('Key')}")
+                    yield (
+                        self.CloudPath(f"s3://{cloud_path.bucket}/{result_key.get('Key')}"),
+                        False,
+                    )
```
jayqi, line 205: @pjbull: Reviewing @GeorgeSabu's #206 (thanks!) made me realize a problem here. I think we need to both check whether `result_key` is size 0 and ends with trailing `/`, similar to what was done in #190.

jayqi (changes requested):

cloudpathlib/s3/s3client.py
pjbull: Yeah, I wanted to keep fixing that issue separate, so I'm going to merge this and then rebase #206 onto it and test with some "fake" directories.

PR #204, opened 2022-02-05
fix: download_to function tried to download path files

yuriolive:
In download directory cases, the `download_to` function was running in an infinite loop. For example:

We have a s3 example bucket with 1 one file in it:
`s3://example/test_file.txt`

The function `iterdir()` will return:

```
s3://example/
s3://example/test_file.txt
````

In this case the `rel_dest` variable will be an empty string for the first case and `test_file.txt` for the second one.

The `f.download_to(destination / rel_dest)` will call the `download_to` function with `re_dest` as empty, causing an infinite loop.

The solution was to just verify if `re_dest` is not empty.

I also added the `parents=True` to `destination.mkdir` so it will create the directory even if is missing intermediate parent directories.

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -587,22 +587,21 @@ class CloudPath(metaclass=CloudPathMeta):
         return self._dispatch_to_local_cache_path("read_text")
 
     # ===========  public cloud methods, not in pathlib ===============
-    def download_to(self, destination: Union[str, os.PathLike]) -> Path:
+    def download_to(self, destination: Union[str, os.PathLike], parents=False) -> Path:
         destination = Path(destination)
         if self.is_file():
             if destination.is_dir():
                 destination = destination / self.name
             return self.client._download_file(self, destination)
         else:
-            destination.mkdir(parents=True, exist_ok=True)
+            destination.mkdir(parents=parents, exist_ok=True)
             for f in self.iterdir():
                 rel = str(self)
                 if not rel.endswith("/"):
                     rel = rel + "/"
 
                 rel_dest = str(f)[len(rel) :]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cloudpathlib/cloudpath.py [hunk 1]
```diff
@@ -594,14 +594,15 @@ def download_to(self, destination: Union[str, os.PathLike]) -> Path:
...
-            destination.mkdir(exist_ok=True)
+            destination.mkdir(parents=True, exist_ok=True)
             for f in self.iterdir():
                 rel = str(self)
                 if not rel.endswith("/"):
                     rel = rel + "/"
 
                 rel_dest = str(f)[len(rel) :]
-                f.download_to(destination / rel_dest)
+                if rel_dest != "":
```
jayqi, line 604: The root cause of this is that `iterdir` improperly yields itself (issue #15), which we have a fix for in #202 that should be merged and released soon.  

This makes the check here redundant. I don't immediately have a strong leaning on whether having the redundant check would be good or bad. Either way, it would also be useful is to add a test for this case.

cloudpathlib/cloudpath.py
```diff
@@ -594,14 +594,15 @@ def download_to(self, destination: Union[str, os.PathLike]) -> Path:
                 destination = destination / self.name
             return self.client._download_file(self, destination)
         else:
-            destination.mkdir(exist_ok=True)
+            destination.mkdir(parents=True, exist_ok=True)
```
jayqi, line 597: I'm not totally bought in on automatically creating missing parents by default. I believe most things (unix programs and Python libraries) do not normally do this. 

A related consideration here is to maybe add a `parents` parameter to `download_to`?

pjbull:
Stale, closing this.

Issue #203, opened 2022-02-05
Make test suite pass with pytest>=7.0


pjbull:
In #202 we added a dev requirement to use pytest<7.0 since the test suite was breaking with the latest pytest version.

Looks like a compatibility issue with `pytest-cases`:
https://github.com/smarie/python-pytest-cases/issues/251

Here's the stack trace:
```
__________________________________________________________________________________________________________________________________________________________________ ERROR collecting tests/test_s3_specific.py __________________________________________________________________________________________________________________________________________________________________
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/pluggy/_hooks.py:265: in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/pluggy/_manager.py:80: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/_pytest/python.py:257: in pytest_pycollect_makeitem
    res = list(collector._genfunctions(name, obj))
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/_pytest/python.py:487: in _genfunctions
    fixtures.add_funcarg_pseudo_fixture_def(self, metafunc, fm)
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/_pytest/fixtures.py:164: in add_funcarg_pseudo_fixture_def
    if not metafunc._calls[0].funcargs:
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/pytest_cases/plugin.py:931: in __getitem__
    return self.calls_list[item]
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/pytest_cases/plugin.py:942: in calls_list
    self.create_call_list_from_pending_parametrizations()
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/pytest_cases/plugin.py:994: in create_call_list_from_pending_parametrizations
    remove_empty_ids(callspec)
../../miniconda3/envs/cloudpathlib/lib/python3.8/site-packages/pytest_cases/fixture_parametrize_plus.py:556: in remove_empty_ids
    callspec._idlist = [c for c in callspec._idlist if not c.startswith(EMPTY_ID)]
[...]

Issue #148, opened 2021-06-29
Ceph compatibility - is_dir() fails for directories when "/" included at end of path


remi-braun:
Hello,

For directories that have a suffix such as .SAFE directories for [Sentinel satellites](https://www.copernicus.eu/fr/propos-de-copernicus/infrastructure/decouvrez-nos-satellites) data, `pathlib` recognizes the directories correctly, although `cloudpathlib` fails:

```python
>>> # On disk
>>> from cloudpathlib import AnyPath
>>> A = AnyPath(r"D:\S2B_MSIL2A_20200114T065229_N0213_R020_T40REQ_20200114T094749.SAFE")
>>> A.is_file()
False
>>> A.is_dir()
True
```

```python
>>> # In th Cloud
>>> client = S3Client(
    endpoint_url=f"https://{AWS_S3_ENDPOINT}",
    aws_access_key_id=os.getenv(AWS_ACCESS_KEY_ID),
    aws_secret_access_key=os.getenv(AWS_SECRET_ACCESS_KEY),
)
>>> client.set_as_default_client()
>>> B = AnyPath("s3://my-bucket/S2B_MSIL2A_20200114T065229_N0213_R020_T40REQ_20200114T094749.SAFE/")
>>> B.is_file()
True
>>> B.is_dir()
False
```

However, it works when removing the final `\` !

```python
>>> # In th Cloud
>>> client = S3Client(
    endpoint_url=f"https://{AWS_S3_ENDPOINT}",
    aws_access_key_id=os.getenv(AWS_ACCESS_KEY_ID),
    aws_secret_access_key=os.getenv(AWS_SECRET_ACCESS_KEY),
)
>>> client.set_as_default_client()
>>> B = AnyPath("s3://my-bucket/S2B_MSIL2A_20200114T065229_N0213_R020_T40REQ_20200114T094749.SAFE")
>>> B.is_file()
False
>>> B.is_dir()
True
```

pjbull:
Hi @remi-braun, I don't reproduce this based on what you specified. Here's an example with a public asset on S3:

```python
>>> file = S3Path('s3://drivendata-public-assets/nested_dir/test.dot/a_file.txt')
>>> file.is_file()
True
>>> file.is_dir()
False
>>> directory = S3Path('s3://drivendata-public-assets/nested_dir/test.dot/')
>>> directory.is_file()
False
>>> directory.is_dir()
True
```

A few questions: 
[...]

PR #206, opened 2022-02-06
s3 fake directory fix

GeorgeSabu:
- Have fixed the fake s3 directory bug 
- Have fixed the test cases
- Added some content to the file in the assets folder in test so that file size is not zero
- Also fixed some warning that are coming while running test cases

### DIFF
--- a/cloudpathlib/client.py
+++ b/cloudpathlib/client.py
@@ -37,7 +37,7 @@ class Client(abc.ABC, Generic[BoundedCloudPath]):
 
     def __del__(self) -> None:
         # make sure temp is cleaned up if we created it
-        if self._cache_tmp_dir is not None:
+        if hasattr(self, "_cache_tmp_dir") and self._cache_tmp_dir is not None:
             self._cache_tmp_dir.cleanup()
 
     @classmethod
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -172,7 +172,7 @@ class CloudPath(metaclass=CloudPathMeta):
 
     def __del__(self):
         # make sure that file handle to local path is closed
-        if self._handle is not None:
+        if hasattr(self, "_handle") and self._handle is not None:
             self._handle.close()
 
     @property
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -197,7 +197,7 @@ class S3Client(Client):
 
                 # files in the directory
                 for result_key in result.get("Contents", []):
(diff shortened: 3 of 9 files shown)


GeorgeSabu:
@pjbull I have updated `cloudpathlib/client.py` and `cloudpathlib/cloudpath.py` as they caused warning while running test cases.

GeorgeSabu:
@pjbull There is a ` attr.exceptions.FrozenInstanceError` that is coming while running test because of which tests are failing which I am not able to reproduce locally.

jayqi:
@GeorgeSabu the error is from https://github.com/drivendataorg/cloudpathlib/issues/203

GeorgeSabu:
@jayqi oops didnt know about it. I did an empty commit to rerun the test to make sure its not some other thing. Thanks for the info.

jayqi:
Another thing to note here is that we've made some changes to `_list_dir` in #202 so there are going to be some merge conflicts.

---

Additionally, some comments about the changes:

- I'm kind of confused about how this fix solves the problem exactly. I believe that this is related to #198? Where is the recursion error coming from, and why is _list_dir the underlying cause? I'd like to see a test that reproduces the recursion error so I can understand why this change would fix it. I think perhaps #198 is not fully explained. 
  - Comparing the discussion from #148 that motivated #198 to the recent discussion in #204, it kind of seems like this isn't maybe a problem with fake directories and actually a general bug with trying to apply `download_to()` on a directory due to how `iterdir()` works. In that case, I think the change proposed in this PR is the wrong fix. #202 should be fixing the underlying cause and we expect to merge it soon.
- I'm also in general not sure the proposed change is exactly the desired behavior. `_list_dir` should yield directories if there are any. However, I think this is perhaps pointing out a weakness in #202 where we don't identify fake directories as directories.

GeorgeSabu:
@jayqi As far as I understood the code the function `iterdir`  which wraps the `_list_dir`. And the `iterdir` is been used as an iterator in download_to as well as `copytree` function. The reason it goes to recursion is that when S3 creates a fake directory what I understands it does is making a PUT request to "Key" `test-folder/`.
```
import boto3

client = boto3.client('s3')

response = client.put_object(
        Bucket='my-top-level-bucket',
        Body='',
        Key='test-folder/'
        )
```
 Now 'test-folder/' behaves likes an object than a prefix. And as `copytree` calls `copytree` internally for its subdirectory causes for it to go into recursion. And same is the case with `download_to` function. 

By checking the Size > 0 I am making sure that this recursion does not happen. I hope this clears.

PR #207, opened 2022-02-06
s3 fake directory fix

pjbull:
Pulling the work from #206 over and rebasing onto the latest master.

---------------------

* [WIP]: taking care of the corner case folder created from S3

* Fix format issues

* [WIP]:updated test case for the s3 file directory test

* removed the linting issue

* Empty commit to rerun the test cases

Co-authored-by: Sabu George <sabugeorge@elucidata-sabs.local>

### DIFF
--- a/cloudpathlib/client.py
+++ b/cloudpathlib/client.py
@@ -37,7 +37,7 @@ class Client(abc.ABC, Generic[BoundedCloudPath]):
 
     def __del__(self) -> None:
         # make sure temp is cleaned up if we created it
-        if self._cache_tmp_dir is not None:
+        if hasattr(self, "_cache_tmp_dir") and self._cache_tmp_dir is not None:
             self._cache_tmp_dir.cleanup()
 
     @classmethod
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -173,7 +173,7 @@ class CloudPath(metaclass=CloudPathMeta):
 
     def __del__(self):
         # make sure that file handle to local path is closed
-        if self._handle is not None:
+        if hasattr(self, "_handle") and self._handle is not None:
             self._handle.close()
 
     @property
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -199,10 +199,11 @@ class S3Client(Client):
 
                 # files in the directory
                 for result_key in result.get("Contents", []):
-                    yield (
-                        self.CloudPath(f"s3://{cloud_path.bucket}/{result_key.get('Key')}"),
-                        False,
-                    )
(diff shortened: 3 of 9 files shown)


cloudpathlib/client.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, local_cache_dir: Optional[Union[str, os.PathLike]] = None):
 
     def __del__(self) -> None:
         # make sure temp is cleaned up if we created it
-        if self._cache_tmp_dir is not None:
+        if hasattr(self, "_cache_tmp_dir") and self._cache_tmp_dir is not None:
```
jayqi, line 40:
```suggestion
        if getattr(self, "_cache_tmp_dir", None) is not None:
```

Should also work and be a little more concise, I think?

cloudpathlib/cloudpath.py
```diff
@@ -173,7 +173,7 @@ def __init__(self, cloud_path: Union[str, "CloudPath"], client: Optional["Client
 
     def __del__(self):
         # make sure that file handle to local path is closed
-        if self._handle is not None:
+        if hasattr(self, "_handle") and self._handle is not None:
```
jayqi, line 176:
```suggestion
        if getattr(self, "_handle"None) is not None:
```

cloudpathlib/s3/s3client.py
```diff
@@ -199,10 +199,11 @@ def _list_dir(self, cloud_path: S3Path, recursive=False) -> Iterable[Tuple[S3Pat
...
                 for result_key in result.get("Contents", []):
-                    yield (
-                        self.CloudPath(f"s3://{cloud_path.bucket}/{result_key.get('Key')}"),
-                        False,
-                    )
+                    if result_key.get("Size") > 0:
+                        yield (
+                            self.CloudPath(f"s3://{cloud_path.bucket}/{result_key.get('Key')}"),
+                            False
+                        )
```
jayqi, line 206: So I think what we want here is a

```python
if result_key.get("Key").endswith("/") and result_key.get("Size") == 0:
    yield (..., True)
else:
    yield (..., False)
```

This way, we cover both empty files and fake directories. 

However, I think we may want to use the `yielded_dirs` as with the non-recursive case, to avoid returning duplicates if the fake directory came up from having children.

tests/assets/dir_0/file0_0.txt
```diff
@@ -0,0 +1 @@
+sample data of file0_0.txt
```
jayqi, line 1: I don't think we should to add content to these to make them non-empty.

jayqi (changes requested):
I think we should also write some tests that involve I/O with some fake directories. This is kind of a blind spot right now.

pjbull:
Superseded by #302

PR #219, opened 2022-04-05
Don't package tests

psarka:


setup.py
```diff
@@ -53,7 +53,7 @@ def load_requirements(path: Path):
     long_description_content_type="text/markdown",
     include_package_data=True,
     name="cloudpathlib",
-    packages=find_packages(),
+    packages=["cloudpathlib"],
```
jayqi, line 56: Hi @psarka, thank you for the PR.

I believe this is not quite right as it excludes submodules. This is what I get from a built distribution using your branch:

```
❯ tree cloudpathlib-0.7.0
cloudpathlib-0.7.0
├── LICENSE
├── MANIFEST.in
├── PKG-INFO
├── README.md
├── cloudpathlib
│   ├── __init__.py
│   ├── anypath.py
│   ├── client.py
│   ├── cloudpath.py
│   └── exceptions.py
├── cloudpathlib.egg-info
│   ├── PKG-INFO
│   ├── SOURCES.txt
│   ├── dependency_links.txt
│   ├── requires.txt
│   └── top_level.txt
├── pyproject.toml
├── requirements.txt
├── setup.cfg
└── setup.py
``` 

I think we actually do still want to use `find_packages` but use arguments to limit down to the modules we want, like
```python
packages=find_packages(include=["cloudpathlib", "cloudpathlib.*"])
```

jayqi (changes requested):

psarka:
Ugh, thanks for catching it.

jayqi (approved):
Looks good, thank you!

PR #220, opened 2022-04-18
Fix error in AnyPath docs

karolzlot:
btw. link to edit on this page doesn't work: https://cloudpathlib.drivendata.org/stable/anypath-polymorphism/


jayqi:
Thanks for fixing! Also opened #221 regarding the broken links.

PR #226, opened 2022-05-16
Implement content type propagation

### LINKED ISSUE #222: S3Path does not set object mime-type when uploading a file on S3
In order to serve correctly S3 object from cloudfront origin, we need to set correctly S3 object mime-type.
Unfortunately the S3 client does not provide any mime-type when calling upload_file.

It could be possible to use at least the mimetypes module for the most common mime-types based on file extension in s3client.py:

    def _upload_file(self, local_path: Union[str, os.PathLike], cloud_path: S3Path) -> S3Path:
        obj = self.s3.Object(cloud_path.bucket, cloud_path.key)

        extra_args = {}
        content_type, content_encoding = mimetypes.guess_type(str(local_path))
        if content_type is not None:
            extra_args["ContentType"] = content_type
        if content_encoding is not None:
            extra_args["ContentEncoding"] = content_encoding

        obj.upload_file(str(local_path), Config=self.boto3_transfer_config, ExtraArgs=extra_args)
        return cloud_path

pjbull:
Implements propagating content types to all of the cloud providers. Does so by adding a kwarg to all clients `content_type_method` that by default is set to the built-in `mimetypes.guess_type`.

Includes tests and documentation.

Changes looks bigger than it is because all of our Local testing resources and Mocks had to be updated to track content types that were set so that we could test them.

Closes #222


jayqi (changes requested):

pjbull:
Thanks @jayqi for review! Integrated all your comments.

HISTORY.md
```diff
@@ -1,8 +1,9 @@
 # cloudpathlib Changelog
 
-## v0.7.2 (UNRELEASED)
+## v0.8.0 (UNRELEASED)
```
jayqi, line 3:
```suggestion
## v0.8.0 (2022-05-19)
```

jayqi (approved):
Looks good, aside from small typo. I think you can bump the version in setup.py and then release after merging.

Issue #235, opened 2022-06-02
Generate Presigned URLs


kabirkhan:
## Overview

It's nice to be able to generate presigned URLs for download/upload so you don't need to worry about passing around auth credentials. 
It's often a little non-obvious how to do these with the downstream clients and each client handles the operation pretty differently so this feels like something that could fit nicely into cloudpathlib.

## Proposed Interface/ implementation

I'd imagine the code for this looking something like the below implementation. The S3Path implementation works and is tested, the others (GSPath/AzureBlobPath) aren't actually tested but are about what the code would look like.

```python

from datetime import datetime, timedelta

class CloudPath:
    def generate_presigned_url(self, expire_seconds: int = 300):
        raise NotImplementedError


class S3Path(CloudPath):
    def generate_presigned_url(self, expire_seconds: int = 300):
        object_key = "/".join(list(self.parts)[2:]) # Everything after the bucket name
        url = self.client.client.generate_presigned_url(
            "get_object",
            Params={"Bucket": self.bucket, "Key": object_key},
            ExpiresIn=expire_seconds,
        )


class GSPath(CloudPath):
    def generate_presigned_url(self, expire_seconds: int = 300):
        object_key = "/".join(list(self.parts)[2:]) # Everything after the bucket name
        gs_client = path.client.client
        creds = gs_client.credentials
        gs_bucket = path.client.client.get_bucket(path.bucket)
        gs_blob = gs_bucket.blob(object_key)
        url = gs_blob.generate_signed_url(
            version="v4",
            expiration=timedelta(seconds=expire_seconds),
            service_account_email=creds.service_account_email,
            access_token=creds.token,
            method="GET"
        )
        return url



from azure.storage.blob import ResourceTypes, BlobSasPermissions, generate_blob_sas


class AzureBlobPath(CloudPath):
    def generate_presigned_url(self, expire_seconds: int = 300):
        object_key = "/".join(list(self.parts)[2:]) # Everything after the bucket name
        az_client = self.client.client
        sas_token = generate_blob_sas(
            az_client.account_name,
            container_name=self.container,
            blob_name=object_key,
            account_key=az_client.credential.account_key,
            permission=BlobSasPermissions(read=True),
            expiry=datetime.utcnow() + timedelta(seconds=expire_seconds)
        )
        url = f"https://{az_client.account_name}.blob.core.windows.net/{self.container}/{object_key}?{sas_token}"
        return url
```

[...]

PR #236, opened 2022-06-03
Add `as_url` method to each CloudPath implementation with ability to generate presigned_urls

kabirkhan:
## Overview

Adds an `as_url` method that returns the `https://` address of the object or generates a presigned URL if `presign=True` is passed.

Based on https://github.com/drivendataorg/cloudpathlib/issues/235

## Usage

```python

path = CloudPath("az://tests/test.txt")

print(path.as_url())
# https://myaccount.blob.core.windows.net/tests/text.txt

print(path.as_url(presign=True))
# https://myaccount.blob.core.windows.net/tests/text.txt?se=...[OTHER_SAS_TOKEN_PARAMS]'

print(path.as_url(presign=True, expire_seconds=10))
# URL will now expire in 10 seconds
# https://myaccount.blob.core.windows.net/tests/text.txt?se=...[OTHER_SAS_TOKEN_PARAMS]'

```

### 1. What should we call this method? `as_url` would be fine but there's already the Pathlib `as_uri` method so it's probably confusing.

Maybe we make `as_uri` return the default `https://` address for the blob instead of the str of the Path. Then we have this method for generating presigned urls be more specific (e.g. `get_signed_url`)
So for the above example:

```python

path = CloudPath("az://tests/test.txt")

print(path.as_uri())
# https://myaccount.blob.core.windows.net/tests/text.txt
# 
# Note: Currently this would return the str of Path so:
# "az://tests/text.txt"

print(path.get_signed_url())
# https://myaccount.blob.core.windows.net/tests/text.txt?se=...[OTHER_SAS_TOKEN_PARAMS]'
```


pjbull:
@kabirkhan Thanks for this, I think the implementation is looking great. A few questions:

(1)  Do you know if any of the provider specific SDKs support getting the URL? Might be more reliable than the f-string version, even in the non-presigned case
(2) The S3 will need to handle if a custom endpoint url is set (e.g., on a MinIO server)
(3) I'm partial to using `as_url` and having that support the presigning, and also leaving `as_uri` alone. I do definitely see the point you brought up that it is potentially confusing, but IMO at least for S3 there's enough treatment of the `s3://...` form as a URI that having both options may provide utility (even if you can get the `as_uri` with `str` already).

Other than those qs, this will need:
 - [ ] tests
 - [ ] documentation updates

Also, do you have creds that you can use to test the various live versions of the services? If not, I can share some with you.

pjbull:
@kabirkhan I know this is a little old, but I'm pushing to get some valuable stale PRs in if possible. Would you have time to wrap the implementation here with the items above?

kabirkhan:
Hey @pjbull, I found some time! lol sorry for the super late reply here. I ended up not needing this for my work but it popped in my head that I never finished this and it's legitimately a useful addition. 

Finished up the main code changes and this now uses the SDK of each provider to get the public URLs as well so local emulators should work.

Started working on the tests as well but not all the there yet.

pjbull:
Awesome, thanks @kabirkhan! I actually had just had to reach into the `S3Client` and do this with the boto client directly, so excited to get this in.

I'll take a look this weekend.

pjbull:
Thanks @kabirkhan! Moving to local branch to run the live backend tests

PR #237, opened 2022-06-03
Fix intersphinx inventory

### LINKED ISSUE #211: Fix API Reference header formatting for Intersphinx
It would be nice if the documentation published an intersphinx `objects.inv` file so that other projects
can use it for cross linking in their documentation.

This only will require adding `enable_inventory` to the `mkdocstrings` configuration in `mkdocs.yml`:

```yaml
  - mkdocstrings:
      # generate objects.inv sphinx inventory file in site/ dir to support external cross references
      enable_inventory: true
```

See https://mkdocstrings.github.io/usage/#cross-references-to-other-projects-inventories

jayqi:
Fixes missing entries from Intersphinx inventory file by including root object in TOC for mkdocstrings. Closes #211.

The inventory file is available at `<base-url>/objects.inv` and can be parsed by [sphobjinv](https://github.com/bskinn/sphobjinv).

Some additional changes:
- Migrates use of `watch` per mkdocstrings [deprecation note](https://mkdocstrings.github.io/usage/#watch-directories)
- Changes mkdocstrings handler from `python` to `python-legacy` to recover old behavior. There are a few issues like https://github.com/mkdocstrings/python/issues/9 that are not yet resolved in the new handler.


pjbull:
Looks good to me! I diffed the old and the new and see the class level ones mentioned in the issue showing up:
image

pjbull:
@jayqi shall we cut a release with this one?

pjbull (approved):

PR #244, opened 2022-08-04
Make prefix not being recognised as dirs anymore

### LINKED ISSUE #208: S3Path.exists() returns True on partial matches
S3Path.exists() is returning True for anything that begins with the same as a real file. For instance, if we have a file named `foobar`, we get a false positive when querying for `foo`.

```
>>> S3Path('s3://my-bucket/foobar').exists()  # This file is really there
True
>>> S3Path('s3://my-bucket/foo').exists()  # This file does NOT exist
True
```

frndmg:
Fixes #208

### DIFF
--- a/tests/test_s3_specific.py
+++ b/tests/test_s3_specific.py
@@ -164,6 +164,36 @@ def test_fake_directories(s3_like_rig):
         assert not test_case.is_file()
 
 
+def test_directories(s3_like_rig):
+    """Fixing bug that marks as directories prefixes of existing paths
+
+    Ref: https://github.com/drivendataorg/cloudpathlib/issues/208
+    """
+    boto3_s3_client = s3_like_rig.client_class._default_client.client
+    if not s3_like_rig.live_server:
+        pytest.skip("This test only runs against live servers.")
+
+    response = boto3_s3_client.put_object(
+        Bucket=f"{s3_like_rig.drive}",
+        Body="123",
+        Key=f"{s3_like_rig.test_dir}/superfile",
+    )
+
+    assert response["ResponseMetadata"]["HTTPStatusCode"] == 200
+
+    super_path = s3_like_rig.create_cloud_path("superfile")
+    assert super_path.exists()
+    assert not super_path.is_dir()
+
+    super_path = s3_like_rig.create_cloud_path("super2")
+    assert not super_path.exists()
+    assert not super_path.is_dir()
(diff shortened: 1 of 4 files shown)


tests/test_s3_specific.py
```diff
@@ -164,6 +164,36 @@ def test_fake_directories(s3_like_rig):
         assert not test_case.is_file()
 
 
+def test_directories(s3_like_rig):
+    """Fixing bug that marks as directories prefixes of existing paths
+
+    Ref: https://github.com/drivendataorg/cloudpathlib/issues/208
+    """
+    boto3_s3_client = s3_like_rig.client_class._default_client.client
+    if not s3_like_rig.live_server:
```
pjbull, line 173: I think we can/should run this against the mocked backend too.

tests/test_s3_specific.py
```diff
@@ -164,6 +164,36 @@ def test_fake_directories(s3_like_rig):
...
+        pytest.skip("This test only runs against live servers.")
+
+    response = boto3_s3_client.put_object(
+        Bucket=f"{s3_like_rig.drive}",
+        Body="123",
+        Key=f"{s3_like_rig.test_dir}/superfile",
+    )
+
+    assert response["ResponseMetadata"]["HTTPStatusCode"] == 200
```
pjbull, line 182: I think that we can just use one of the files that gets populated by the rig automatically (for example, like we do here):
https://github.com/drivendataorg/cloudpathlib/blob/2e52c0515547e998e0b7c84e667627df4ac332b6/tests/test_cloudpath_file_io.py#L17-L19

pjbull (changes requested):
Thanks @frndmg! We have to get this onto a branch rather than a fork to do the live tests, but lets make the updates here for the tests first.

pjbull:
Thanks @frndmg! Once CI passes here, I'll pull over to a local branch to run the tests against live servers

pjbull:
Actually, would you mind adding a note in the HISTORY.md as well? I believe you need to start a new "unreleased" section. Here's an old one for reference:
https://github.com/drivendataorg/cloudpathlib/blob/85268c8470f7aa279691eaa96caaaa8652d20204/HISTORY.md?plain=1#L3-L5

frndmg:
@pjbull feel free to update the message in the `HISTORY.md` if you want

PR #245, opened 2022-08-11
Make prefix not being recognised as dirs anymore (#244)

### LINKED ISSUE #208: S3Path.exists() returns True on partial matches

pjbull:
Branch for running live server tests for PR #244

Closes #208

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,9 @@
 # cloudpathlib Changelog
 
+## v0.9.1 (UNRELEASED)
+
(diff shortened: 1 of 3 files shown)


pjbull:
Thanks @frndmg!

Issue #246, opened 2022-08-15
AnyPath is not a subclass of AnyPath


chbehrens:
It would be helpful if issubclass(AnyPath, AnyPath) returns True as usually expected if a class is checked against itself. Do you see any possibility to modify it in such a way or is there a good reason why this shouldn't be the case? Unfortunately, I'm not familiar enough with ABCs to directly suggest a solution.

I came across a use case where I want to make sure, that the type of a field is a subclass of a list of classes. Originally, I used pathlib.Path there and checked against (str, numbers.Number, Path) but as cloud paths should become an option, I'd like to replace pathlib.Path with cloudlib.AnyPath.

pjbull:
Closed by #251

PR #247, opened 2022-08-16
Strictly follow PEP3119 for AnyPath __subclasscheck__ implementation

chbehrens:
The currently implemented subclass check for the AnyPathMeta class results in issubclass(AnyPath, AnyPath) returning False, opposed its expected behavior. With this PR I suggest to switch to the exact implementation shown in [PEP 3119](https://peps.python.org/pep-3119/#overloading-isinstance-and-issubclass). It should resolve this problem while preserving the intended behavior with respect to CloudPath and pathlib.Path.

See https://github.com/drivendataorg/cloudpathlib/issues/246

### DIFF
--- a/cloudpathlib/anypath.py
+++ b/cloudpathlib/anypath.py
@@ -12,10 +12,11 @@ class AnyPathMeta(type):
     [PEP 3119](https://www.python.org/dev/peps/pep-3119/#overloading-isinstance-and-issubclass)."""
 
     def __instancecheck__(cls, inst):
-        return isinstance(inst, CloudPath) or isinstance(inst, Path)
+        return any(cls.__subclasscheck__(c) for c in {type(inst), inst.__class__})
 
     def __subclasscheck__(cls, sub):
-        return issubclass(sub, CloudPath) or issubclass(sub, Path)
+        candidates = cls.__dict__.get("__subclass__", set()) | {cls}
+        return any(c in candidates for c in sub.mro())
 
 
 class AnyPath(metaclass=AnyPathMeta):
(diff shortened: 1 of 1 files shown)


chbehrens:
Sorry, I tested it before but messed it up during clean-up. Should be complete now.

jayqi:
I'm wondering if my original decision to use custom `__subclasscheck__` and `__instancecheck__` is the wrong approach here. 

I believe another way of accomplishing what we want (that is also part of PEP 3119) is to have `AnyPath` subclass `abc.ABC` and use [`ABC.register`](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register). This would mean getting rid of our `AnyPathMeta` class since we wouldn't need it, and instead having 

```python
AnyPath.register(Path)
AnyPath.register(CloudPath)
```

I just did a little bit of light testing, and I think it should work equivalently and also resolve the issue of `issubclass(AnyPath, AnyPath)`. 

---

I'm lukewarm on the idea of copying the exact implementation for `__subclasscheck__` and `__instancecheck__` from PEP 3119. PEP 3119 describes those as "naively simple" from-scratch implementations. According to later in PEP 3119:

> The call isinstance(x, C) first checks whether C.__instancecheck__ exists, and if so, calls C.__instancecheck__(x) instead of its normal implementation. Similarly, the call issubclass(D, C) first checks whether C.__subclasscheck__ exists, and if so, calls C.__subclasscheck__(D) instead of its normal implementation.

So that means by doing it this way with `__subclasscheck__` and `__instancecheck__`, we're entirely responsible for the logic of `isinstance` and `issubclass` rather than leaning on any default logic.

pjbull:
I like the implementation @jayqi laid out. @chbehrens would you be willing to take a hack at that version? 

Could you also add a test to the subclass tests to check the bug you filed in #246?

pjbull:
Resolved by #251

PR #249, opened 2022-08-17
Skip docstrings if running with `-OO`

### LINKED ISSUE #238: Module fails to import when Python is ran with -OO
Since `-OO` removes docstrings, the module fails on import as shown in the following trace:

```python
[...]
    from cloudpathlib import AnyPath, CloudPath
  File "/home/javier/.virtualenvs/WebsiteTopicClassifier/lib/python3.6/site-packages/cloudpathlib/__init__.py", line 3, in <module>
    from .anypath import AnyPath
  File "/home/javier/.virtualenvs/WebsiteTopicClassifier/lib/python3.6/site-packages/cloudpathlib/anypath.py", line 5, in <module>
    from .cloudpath import InvalidPrefixError, CloudPath
  File "/home/javier/.virtualenvs/WebsiteTopicClassifier/lib/python3.6/site-packages/cloudpathlib/cloudpath.py", line 136, in <module>
    class CloudPath(metaclass=CloudPathMeta):
  File "/home/javier/.virtualenvs/WebsiteTopicClassifier/lib/python3.6/site-packages/cloudpathlib/cloudpath.py", line 127, in __init__
    docstring = getattr(Path, attr).__doc__ + " _(Docstring copied from pathlib.Path)_"
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

pjbull:
Manually tested and confirmed that PyPI version fails and this version succeeds.

Fixes #238

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -123,6 +123,7 @@ class CloudPathMeta(abc.ABCMeta):
                 not attr.startswith("_")
                 and hasattr(Path, attr)
                 and hasattr(getattr(Path, attr), "__doc__")
+                and getattr(getattr(Path, attr), "__doc__") is not None
             ):
                 docstring = getattr(Path, attr).__doc__ + " _(Docstring copied from pathlib.Path)_"
                 getattr(cls, attr).__doc__ = docstring
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -122,8 +122,7 @@ class CloudPathMeta(abc.ABCMeta):
             if (
                 not attr.startswith("_")
                 and hasattr(Path, attr)
-                and hasattr(getattr(Path, attr), "__doc__")
-                and getattr(getattr(Path, attr), "__doc__") is not None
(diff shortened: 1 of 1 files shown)


cloudpathlib/cloudpath.py
```diff
@@ -123,6 +123,7 @@ def __init__(cls, name, bases, dic):
                 not attr.startswith("_")
                 and hasattr(Path, attr)
                 and hasattr(getattr(Path, attr), "__doc__")
+                and getattr(getattr(Path, attr), "__doc__") is not None
```
jayqi, line 126: Couldn't both of these lines (125 and 126) just be simplified to `and getattr(getattr(Path, attr), "__doc__")`? This would both check if it exists, and that it casts as a boolean to positive (would exclude the `None` case). Not a big deal.

jayqi (approved):

cloudpathlib/cloudpath.py
pjbull: Yep, good call. That should work as long as we add a [default to `getattr`](https://docs.python.org/3/library/functions.html#getattr) so it doesn't raise.

PR #251, opened 2022-08-17
Use abc.ABC as baseclass for AnyPath to fix issubclass behavior

chbehrens:
Following @jayqi's suggestion in https://github.com/drivendataorg/cloudpathlib/pull/247, this PR lets AnyPath inherit from abc.ABC and sets its subclasses via [ABC.register](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register). It also adds a test to check the bug filed in https://github.com/drivendataorg/cloudpathlib/issues/246.

Fixes https://github.com/drivendataorg/cloudpathlib/issues/246

### DIFF
--- a/cloudpathlib/anypath.py
+++ b/cloudpathlib/anypath.py
@@ -44,7 +44,7 @@ class AnyPath(ABC):
         return cls.__new__(cls, value)
 
 
-AnyPath.register(CloudPath)
+AnyPath.register(CloudPath)  # type: ignore
 AnyPath.register(Path)
 
 
--- a/cloudpathlib/anypath.py
+++ b/cloudpathlib/anypath.py
@@ -1,4 +1,5 @@
 import os
+from abc import ABC
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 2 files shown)


pjbull:
Hm @chbehrens, looks like `mypy` and `ABC` have some rough edges. See https://github.com/python/mypy/issues/2922, https://github.com/python/mypy/issues/5374, https://stackoverflow.com/questions/69330256/how-to-get-an-abstract-dataclass-to-pass-mypy

Ironically, it looks like the `metaclass` implementation might pass `mypy` (see https://github.com/python/mypy/issues/1459#issuecomment-216063442)

I'm ok with a  `# type: ignore` on the line that is erroring. That work for you @jayqi?

jayqi:
I'm also fine with a `# type: ignore`. 

I was also looking at https://github.com/python/mypy/issues/5374 and it seems like it's possible this is addressed very recently in mypy (just 5 days ago) by https://github.com/python/mypy/pull/13398. That PR was merged but is not released yet, so I think we'll potentially look at revisiting once the next release of mypy is published.

chbehrens:
Ok, I added the `# type: ignore` and will try to keep it in mind.

pjbull:
Thanks @chbehrens! You can ignore the failure for now, Github Actions is having an outage:
https://www.githubstatus.com/

I'll rerun these later once they stabilize the service

pjbull:
Changes look good, thanks @chbehrens! I'm going to move this to a local branch to run the live server tests.

PR #253, opened 2022-08-17
Add boto3 extra_args field for upload operation

benbemoh:
There are many use cases where uploading unencrypted objects to an Amazon S3 bucket is denied.
For example an S3 bucket policy which enforces the encryption requirement when users upload objects, unless they are using client-side or server-side encryption (see [this aws article](https://aws.amazon.com/blogs/security/how-to-prevent-uploads-of-unencrypted-objects-to-amazon-s3/) for details).

This PR addresses the case of using a server-side encryption when uploading objects to an Amazon S3 bucket.

Below is a python code example to illustrate how to use the new changes:
```python
extra_args = {"ServerSideEncryption": "AES256"}
client = S3Client(boto3_upload_extra_args=extra_args)
```
@pikulmar: Thank you for the suggestions and for reviewing the PR

### DIFF
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -1,7 +1,7 @@
 import mimetypes
 import os
 from pathlib import Path, PurePosixPath
-from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union, Final
 
 
 from ..client import Client, register_client_class
@@ -36,6 +36,7 @@ class S3Client(Client):
         local_cache_dir: Optional[Union[str, os.PathLike]] = None,
         endpoint_url: Optional[str] = None,
         boto3_transfer_config: Optional["TransferConfig"] = None,
+        boto3_upload_extra_args: Optional[Dict[str, str]] = None,
         content_type_method: Optional[Callable] = mimetypes.guess_type,
     ):
         """Class constructor. Sets up a boto3 [`Session`](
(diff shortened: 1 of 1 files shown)


pjbull:
Thanks @benbemoh. Can you file a related issue where discussion of this change can happen? I see the value, but I want to make sure we cover all the potential `ExtraArgs` usages (not just upload) in one go, and we've got a clean API for passing them.

benbemoh:
@pjbull : Thanks for your reply. 
Two use-cases that comes to my mind  which I know that they require also ExtraArgs encryption:
1. When generating a presigned-s3-url for HTTP post method in order to upload file ([generate_presigned_post](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.generate_presigned_post)). I see that there is an related issue opened [here](https://github.com/drivendataorg/cloudpathlib/issues/235).
```python
# Example of code to generate an s3 presigned url for HTTP POST to upload a file 
fields = {"x-amz-server-side-encryption": "AES256"}
conditions = {"x-amz-server-side-encryption": "AES256"}
 info = {
      "Bucket": bucket_name,
      "Key": object_name,
      "Fields": fields,
      "Conditions": conditions,
      "ExpiresIn": expiration,
  }
  try:
      response: Dict = s3_client.generate_presigned_post(**info)
  except ClientError as e:
      extra_info = f"Error in generate_presigned_post with: {info}."
      raise RuntimeError(e, extra_info)
```
2. When copying S3 objects using these two methods [S3.Object.copy_from()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Object.copy_from) and [S3.Object.copy()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Object.copy).
I think the  cloudpathlib code [here](https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/s3/s3client.py#L211) can be updated like this:
```python
def _move_file(self, src: S3Path, dst: S3Path, remove_src: bool = True) -> S3Path:
    # just a touch, so "REPLACE" metadata
    if src == dst:
        o = self.s3.Object(src.bucket, src.key)
        o.copy_from(
            CopySource={"Bucket": src.bucket, "Key": src.key},
            Metadata=self._get_metadata(src).get("extra", {}),
            MetadataDirective="REPLACE",
            ServerSideEncryption=self.boto3_upload_extra_args.get("ServerSideEncryption"),
        )

    else:
        target = self.s3.Object(dst.bucket, dst.key)
        target.copy(
            CopySource={"Bucket": src.bucket, "Key": src.key},
            ExtraArgs=self.boto3_upload_extra_args.copy(),
        )

        if remove_src:
            self._remove(src)
    return dst
```

benbemoh:
@pjbull : Could you please give us a feed-back regarding the two points the mentioned above?

pjbull:
Thanks for the patience @benbemoh. @gaisensei has filed a related issue, #254 where we can discuss this.

PR #257, opened 2022-08-18
Use abc.ABC as baseclass for AnyPath to fix issubclass behavior (#251)

pjbull:
* Use abc.ABC as baseclass for AnyPath

* add type ignore for ABC.register

---------------

Local branch to run live tests


pjbull:
Thanks @chbehrens!

PR #268, opened 2022-08-23
Use TypeVar for register_path_class decorator return type

ringohoffman:
With the current implementation of `register_path_class`:

```python
def register_path_class(key: str):
    def decorator(cls: type):
        if not issubclass(cls, CloudPath):
            raise TypeError("Only subclasses of CloudPath can be registered.")
        global implementation_registry
        implementation_registry[key]._path_class = cls
        cls._cloud_meta = implementation_registry[key]
        return cls

    return decorator
```

the inferred type returned by `decorator` is `Type[CloudPath]` ... meaning we are losing class-specific type hint information when type hinting as subclasses of `CloudPath`:

Screen Shot 2022-08-22 at 11 47 27 PM

A simple fix for this is to explicitly return the same type that the decorator operates on using a `TypeVar`.
Screen Shot 2022-08-22 at 11 47 50 PM

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -12,7 +12,7 @@ from pathlib import (  # type: ignore
     _posix_flavour,
     _PathParents,
 )
-from typing import Any, IO, Iterable, Dict, Optional, TYPE_CHECKING, Union
+from typing import Any, IO, Iterable, Dict, Optional, Type, TYPE_CHECKING, TypeVar, Union
 from urllib.parse import urlparse
 from warnings import warn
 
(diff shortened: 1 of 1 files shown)


ringohoffman:
@pjbull any thoughts?

pjbull:
Change looks good to me, thanks @ringohoffman.

> I updated your workflow to not run the live tests on forks

Could you back this out? Currently, the we use the failing live tests as a signal to maintainers to remember to merge contributions into a repo-local branch first and run the live tests to make sure everything passes on real servers. Not the ideal workflow, but it helps us ensure the main branch is always working against real providers. Hopefully one day we can just add a secondary approval for this that can be triggered after code review.

ringohoffman:
@pjbull thanks for letting me know! I undid the change.

pjbull:
Thanks!

PR #269, opened 2022-09-07
Use TypeVar for register_path_class decorator return type (#268)

pjbull:
Repo-local branch for #268


pjbull:
Thanks @ringohoffman !

PR #275, opened 2022-10-11
Caching file/dir metadata for GSClient

### LINKED ISSUE #176: Speed up `is_dir` on entries returned by `iterdir` and `glob`

Gilthans:
This is an attempt to resolve #176 .
It's a first attempt, I'd mainly like to know if this is what the maintainers had in mind. If it is, I can add some tests - which begs the question, what tests would we like to see?

### DIFF
--- a/cloudpathlib/gs/gsclient.py
+++ b/cloudpathlib/gs/gsclient.py
@@ -1,8 +1,20 @@
+import dataclasses
 from datetime import datetime
 import mimetypes
 import os
 from pathlib import Path, PurePosixPath
-from typing import Any, Callable, Dict, Iterable, Optional, TYPE_CHECKING, Tuple, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    Optional,
+    TYPE_CHECKING,
+    Tuple,
+    Union,
+    MutableMapping,
+)
+from weakref import WeakKeyDictionary
 
 from ..client import Client, register_client_class
 from ..cloudpath import implementation_registry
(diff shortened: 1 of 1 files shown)


Gilthans:
@pjbull friendly ping :)

pjbull:
Thanks @Gilthans, nice work on a first pass on this. Definitely helpful for thinking about the problem.

Caching is a big enough feature/change that we'd prefer to spec it out before thinking about implementation. For example, are we confident we can implement it the same way across backends? What settings/config/env vars do we provide to end users to control behavior? Is this better implemented in the `*Client` or `*Path` objects? What exactly goes in the cache? Is the cache limited in size? Do we provide APIs for clearing the cache?

I think all of these are answerable, but worth making a plan ahead of time to make sure they don't catch us out later and increase the maintenance burden.

Gilthans:
Sure, that does sound reasonable. How does such a discussion happen currently? Is it done strictly by the maintainer team, or is it an open discussion in an issue?
I'd love to get the ball rolling since this feature would be very helpful in our project

pjbull:
@Gilthans We can do this collaboratively in an issue that is opened specifically for metadata caching. Thanks!

PR #276, opened 2022-10-11
Short-circuit 'easy' rglob/glob patterns

### LINKED ISSUE #274: glob and rglob can be slow for large directories or lots of files
Hello!
While gleefully using `cloudpathlib`, I needed a recursive iteration of files in a directory. This directory is large (6238 files), so my first approach - a recursive `iterdir()` + `is_file()` - took waaaay too long (likely due to #176).
I remembered that `glob` made better use of the cloud list calls, so I tried `list(p.rglob("*"))`. In my directory, that took 17m:34s.
I then tried to 'cheat' and call `[f for f, is_dir in p.client._list_dir(p, recursive=True) if not is_dir]`. It took 1.435s.

I looked at the glob logic, but I still can't understand why the discrepency (Using Google Cloud). This may warrant another issue, but 

However, I wonder if it might not be a good idea to make `_list_dir` a public function in the meantime as a workaround.
Another option is to add `recursive` and/or `files_only` keywords to `iterdir`. This deviates from pathlib API, but since these are *added* keywords, it might be OK?

I'm suggesting these options even though solving #176 would probably solve most issues, but these solutions are much simpler.
I'd of course be happy to send a PR.
WDYT?

Gilthans:
Resolves #274

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -361,6 +361,12 @@ class CloudPath(metaclass=CloudPathMeta):
             yield self.client.CloudPath(f"{self.cloud_prefix}{self.drive}{p}")
 
     def glob(self, pattern):
+        if pattern == "*":
+            yield from (path for path, is_file in self.client._list_dir(self, recursive=False))
+            return
+        if pattern == "**/*":
+            yield from (path for path, is_file in self.client._list_dir(self, recursive=True))
+            return
         self._glob_checks(pattern)
 
         pattern_parts = PurePosixPath(pattern).parts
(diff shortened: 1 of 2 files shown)


Gilthans:
@pjbull friendly ping :)

pjbull:
Thanks @Gilthans, while I think this fix looks directionally reasonable, but I'd like to do a bit more investigation on why performance is bad in #274.

I assumed that it was because there was a network call somewhere in the glob path, but we actually only call the list dir once here:
https://github.com/drivendataorg/cloudpathlib/blob/b7f601023e478adbe400a4bbc57efae5bfc95a81/cloudpathlib/cloudpath.py#L349-L358

If you get a chance, could you do some profiling and report in #274? I'm hoping there may be a better more general fix than the short circuit if this isn't networking related (or maybe it is and I just missed where the call is in my scan of the code).

Gilthans:
> I assumed that it was because there was a network call somewhere in the glob path, but we actually only call the list dir once here:

I actually came to the same conclusion before opening the issue. I got that far into the code and couldn't figure out why it was slow either.
I'd love to do this profiling (one of my hobbies actually :) ), but in the meantime, maybe we could push forward with this? It seems to me like a net-positive regardless

Issue #277, opened 2022-10-14
Renaming a file to have the same name causes it to be deleted, error raised


spdjudd:
Hi,

Thanks for your work on cloudpathlib, much appreciated!

I came across this issue when running a script to rename a set of files - one of them happened to have the correct name already, but calling `CloudPath('s3://bucket/folder/filea.pq').rename(CloudPath('s3://bucket/folder/filea.pq'))` causes the file to be lost, and an error raised:

```File ~\anaconda3\lib\site-packages\cloudpathlib\cloudpath.py:451, in CloudPath.rename(self, target)
    448 def rename(self, target: "CloudPath") -> "CloudPath":
    449     # for cloud services replace == rename since we don't just rename,
    450     # we actually move files
--> 451     return self.replace(target)

File ~\anaconda3\lib\site-packages\cloudpathlib\cloudpath.py:445, in CloudPath.replace(self, target)
    442 if target.exists():
    443     target.unlink()
--> 445 self.client._move_file(self, target)
    446 return target

File ~\anaconda3\lib\site-packages\cloudpathlib\s3\s3client.py:217, in S3Client._move_file(self, src, dst, remove_src)
    213 if src == dst:
    214     o = self.s3.Object(src.bucket, src.key)
    215     o.copy_from(
    216         CopySource={"Bucket": src.bucket, "Key": src.key},
--> 217         Metadata=self._get_metadata(src).get("extra", {}),
    218         MetadataDirective="REPLACE",
    219     )
    221 else:
    222     target = self.s3.Object(dst.bucket, dst.key)

...
NoSuchKey: An error occurred (NoSuchKey) when calling the GetObject operation: The object 'folder/filea.pq' was not found in the bucket 'bucket'
```
Seems like checking `target` to see if it's the same path as `self` before unlinking could prevent this - happy to create a PR if that sounds ok?

pjbull:
@spdjudd Yep, that seems right, good catch! Would be happy to accept a PR with that fix and a test added. Thanks.

pjbull:
Closed by #279

Issue #64, opened 2020-09-01
replace and rename don't handle "directories"


jayqi:
`CloudPath.replace` and `CloudPath.rename` (which are the same function for cloudpathlib) do not handle "directories". 

The pathlib docs say:

```
Rename this file or directory to the given target, and return a new Path instance pointing to target. If target points to an existing file or directory, it will be unconditionally replaced.
```

```
>>> from cloudpathlib import CloudPath
>>> s = CloudPath("s3://my-bucket/")
>>> (s / "foo" / "bar.txt").touch()
>>> (s / "foo").replace(s / "baz")
Traceback (most recent call last):
...
botocore.exceptions.ClientError: An error occurred (404) when calling the HeadObject operation: Not Found
```

PR #278, opened 2022-10-19
Prevent file loss/exceptions when renaming

spdjudd:
Fixes issue #277 

# Change Summary
1. Do nothing if renaming a file to have the same path as it currently has - previously the file was unlinked then an exception raised when trying to copy
2. Raise an explicit exception when attempting to rename a directory - the implementation only supports rename/replace of files, which raises an exception from within _move_file() if passed a directory

# Tests
I've added test conditions where it seemed appropriate within the existing test methods to exercise these changes - happy to move to separate test methods if preferred?

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -453,6 +453,15 @@ class CloudPath(metaclass=CloudPathMeta):
                 f"The target based to rename must be an instantiated class of type: {type(self)}"
             )
 
+        if self.is_dir():
+            raise CloudPathIsADirectoryError(
+                f"Path {self} is a directory; rename/replace the files recursively."
+            )
+
+        if target == self:
+            # Request is to replace/rename this with the same path - nothing to do
+            return self
+
         if target.exists():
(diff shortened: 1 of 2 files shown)


cloudpathlib/cloudpath.py
```diff
@@ -453,6 +453,15 @@ def replace(self, target: "CloudPath") -> "CloudPath":
                 f"The target based to rename must be an instantiated class of type: {type(self)}"
             )
 
+        if self.is_dir():
+            raise CloudPathIsADirectoryError(
+                f"Path {self} is a directory; rename/replace the files recursively."
+            )
```
pjbull, line 459: Does pathlib error if you call `replace` on a directory? I assume not. If not, can we add an issue to track implementing `replace` for directories?

pjbull (approved):

cloudpathlib/cloudpath.py
spdjudd: Actually I just checked and I think this is already logged as issue #64

PR #279, opened 2022-10-20
Prevent file loss/exceptions when renaming: (#278)

pjbull:
Local branch of #278 for running live tests

----------

Do nothing if renaming to the same path
Explicit exception when attempting to rename a dir


pjbull:
Thanks @spdjudd !

spdjudd:
No problem, thanks for maintaining a great library!

PR #281, opened 2022-10-24
Add `dirs_exist_ok` to `CloudPath.copytree`

### LINKED ISSUE #144: Add a `dirs_exist_ok` parameter to the `copytree` implementation
We could add a `dirs_exist_ok` parameter to copy tree. To do this, we would also need to raise on `mkdir` if dir exists for cloud implementations.  For most clients `mkdir` is a no-op right now, so that logic would need to be added/tested.

---------

It seems like `shutil.copytree` has a `dirs_exist_ok` parameter that controls this `exist_ok`. Do we want to mimic that interface?

https://docs.python.org/3/library/shutil.html#shutil.copytree

_Originally posted by @jayqi in https://github.com/drivendataorg/cloudpathlib/pull/142#discussion_r634894797_

klwetstone:
closes #144

Adds `dirs_exist_ok` argument to `CloudPath.copytree`

The original issue suggests adding the `dirs_exist_ok` functionality through `mkdir`. Using `mkdir` on `S3Path` doesn't actually create a directory, since we can't make empty directories in cloud storage. Adding functionality to a no-op `mkdir` seemed misleading, so the new argument doesn't use `mkdir` to check if the destionation directory exists. Instead the functionality is added to `copytree` outside of `mkdir` with the more explicit `if destination.exists() and not dirs_exist_ok`.

### Questions

- Since `.mkdir` doesn't work on either `GSPath` or `S3Path`, should we raise an implementation error when it is called? Currently it runs without error, but doesn't actually create a directory.
  - For example, the `destination.mkdir` [line](https://github.com/drivendataorg/cloudpathlib/blob/fa8d9fd33cef12ed693d9e9cd69158b0f50b6d14/cloudpathlib/cloudpath.py#L811) in `CloudPath.copytree` doesn't actually create a cloud directory if it doesn't exist. The directory is created when the first file is copied with [`subpath.copy(...)`](https://github.com/drivendataorg/cloudpathlib/blob/fa8d9fd33cef12ed693d9e9cd69158b0f50b6d14/cloudpathlib/cloudpath.py#L817). Do we need to keep this line for cases when the destination directory is local?
  - If we raise an implementation error instead of pass without doing anything, the new [`S3Path.mkdir`](https://github.com/drivendataorg/cloudpathlib/blob/fa8d9fd33cef12ed693d9e9cd69158b0f50b6d14/cloudpathlib/s3/s3path.py#L41) definition would be:
  ```python
    def mkdir(self, parents=False, exist_ok=False):
        raise CloudPathNotImplementedError(
            "Empties directories cannot be created in cloud storage"
        )
  ```
  - We can add something similar, but with a general `NotImplementedError`, to [`GSPath.mkdir`](https://github.com/drivendataorg/cloudpathlib/blob/fa8d9fd33cef12ed693d9e9cd69158b0f50b6d14/cloudpathlib/gs/gspath.py#L41)

- There is a note in the comments that empty directories can't be created in other forms of cloud storage (`GSPath`) either. It seems like we can handle `GSPath` and `S3Path` similarly for adding `dirs_exist_ok`, is that right?

### DIFF
--- a/tests/test_cloudpath_upload_copy.py
+++ b/tests/test_cloudpath_upload_copy.py
@@ -214,13 +214,17 @@ def test_copytree(rig, tmpdir):
     # cloud dir to local dir that exists
     p = rig.create_cloud_path("dir_1")
     local_out = Path(tmpdir.mkdir("copytree_from_cloud"))
-    p.copytree(local_out)
+    p.copytree(local_out, dirs_exist_ok=True)
     assert assert_mirrored(p, local_out)
 
     # str version of path
     local_out = Path(tmpdir.mkdir("copytree_to_str_path"))
-    p.copytree(str(local_out))
+    p.copytree(str(local_out), dirs_exist_ok=True)
     assert assert_mirrored(p, local_out)
+
+    # test dirs_exist_ok
+    with pytest.raises(CloudPathFileExistsError):
+        p.copytree(str(local_out), dirs_exist_ok=False)
 
     # cloud dir to local dir that does not exist
     local_out = local_out / "new_folder"
@@ -235,14 +239,18 @@ def test_copytree(rig, tmpdir):
     # cloud dir to cloud dir that exists
     p2 = rig.create_cloud_path("new_dir2")
     (p2 / "existing_file.txt").write_text("asdf")  # ensures p2 exists
-    p.copytree(p2)
+    p.copytree(p2, dirs_exist_ok=True)
     assert assert_mirrored(p2, p, check_no_extra=False)
+
+    # test dirs_exist_ok
+    with pytest.raises(CloudPathFileExistsError):
+        p.copytree(p2, dirs_exist_ok=False)
 
     (p / "new_file.txt").write_text("hello!")  # add file so we can assert mirror
     with pytest.raises(OverwriteNewerCloudError):
-        p.copytree(p2)
+        p.copytree(p2, dirs_exist_ok=True)
 
-    p.copytree(p2, force_overwrite_to_cloud=True)
+    p.copytree(p2, force_overwrite_to_cloud=True, dirs_exist_ok=True)
     assert assert_mirrored(p2, p, check_no_extra=False)
 
     # additional files that will be ignored using the ignore argument
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -785,6 +785,7 @@ class CloudPath(metaclass=CloudPathMeta):
             ],
             collections.abc.Iterable,
         ] = None,
+        dirs_exist_ok: bool = False,
     ) -> Union[Path, "CloudPath"]:
         """Copy self to a directory, if self is a directory."""
         if not self.is_dir():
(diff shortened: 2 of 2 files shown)


jayqi:
> The original issue suggests adding the dirs_exist_ok functionality through mkdir. Using mkdir on S3Path doesn't actually create a directory, since we can't make empty directories in cloud storage. Adding functionality to a no-op mkdir seemed misleading, so the new argument doesn't use mkdir to check if the destination directory exists.

This actually surfaces a tricky issue that I hadn't thought of before. Currently, _because_ `mkdir` is implemented as a no-op it doesn't match pathlib's behavior. By default, pathlib's `mkdir` has `exists_ok=False`. Accordingly, the matching behavior should be for `mkdir` to—by default—check if the "directory" "exists" in cloud storage, and error if it already does. It seems like there'd be some performance penalty for this. Thoughts, @pjbull? 

> Since .mkdir doesn't work on either GSPath or S3Path, should we raise an implementation error when it is called? Currently it runs without error, but doesn't actually create a directory.

Definitely not. This is intentionally a no-op so you can copy pathlib's implementation logic to our cloud implementations and not have to make changes to take out all of the `mkdir` calls.

tests/test_cloudpath_upload_copy.py
```diff
@@ -235,14 +239,18 @@ def test_copytree(rig, tmpdir):
...
+    with pytest.raises(CloudPathFileExistsError):
+        p.copytree(p2, dirs_exist_ok=False)
+
     (p / "new_file.txt").write_text("hello!")  # add file so we can assert mirror
     with pytest.raises(OverwriteNewerCloudError):
-        p.copytree(p2)
+        p.copytree(p2, dirs_exist_ok=True)
 
-    p.copytree(p2, force_overwrite_to_cloud=True)
+    p.copytree(p2, force_overwrite_to_cloud=True, dirs_exist_ok=True)
```
jayqi, line 253: Can you add a new test that tests the behavior when `dirs_exist_ok` is False?

jayqi (changes requested):

tests/test_cloudpath_upload_copy.py
klwetstone: Yup! That's added [here](https://github.com/drivendataorg/cloudpathlib/blob/25c4ce7faf1610cc9926d0da8b7430b6686dbf3a/tests/test_cloudpath_upload_copy.py#L225) and [here](https://github.com/drivendataorg/cloudpathlib/blob/25c4ce7faf1610cc9926d0da8b7430b6686dbf3a/tests/test_cloudpath_upload_copy.py#L245).

jayqi: Ah, sorry, didn't see those. 👍

pjbull:
Going through some stale PRs to get them moving.

I think that we shouldn't introduce `dirs_exist_ok=False` as the default without a deprecation cycle. There's too much of a chance people use `copytree` relying on the current behavior. 

I think I'd go with an implementation that looks something like:

```python
def copytree(..., dirs_exist_ok: bool = None) -> Union[Path, "CloudPath"]:
    if dirs_exist_ok is None:
        warnings.warn("In a future version, `CloudPath.copytree` will have a `dirs_exist_ok=False` kwarg. Please specify `dirs_exist_ok=True` in your code if you do not want that behavior (issue #144).", CloudPathDeprecationWarning)
        dirs_exist_ok = True  # match current behavior until dirs_exist_ok=False introduced

    ...
```

Issue #287, opened 2022-11-09
Support for `with_stem`


Gilthans:
Hello!
Today we noticed cloudpathlib doesn't support [`with_stem`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.with_stem) which was added in python 3.9.
What's the policy on this kind of compatibility (i.e. functions that exist on only some of the supported versions of python)?
Would a PR to introduce this be accepted?

pjbull:
This is released as of `0.16.0`, thanks @Gilthans!

PR #290, opened 2022-11-20
Added with_stem support

Gilthans:
#287

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -760,7 +760,10 @@ class CloudPath(metaclass=CloudPathMeta):
         return self._dispatch_to_path("suffixes")
 
     def with_stem(self, stem: str) -> Self:
-        return self._dispatch_to_path("with_stem", stem)
+        try:
+            return self._dispatch_to_path("with_stem", stem)
+        except AttributeError:
+            return self.with_name(stem + self.suffix)
 
     def with_name(self, name: str) -> Self:
         return self._dispatch_to_path("with_name", name)
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -759,6 +759,9 @@ class CloudPath(metaclass=CloudPathMeta):
     def suffixes(self) -> List[str]:
         return self._dispatch_to_path("suffixes")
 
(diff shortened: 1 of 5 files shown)


pjbull:
Thanks @Gilthans, could you add:
 - [ ] test
 - [ ] update methods table (you can run [this script](https://github.com/drivendataorg/cloudpathlib/blob/master/docs/make_support_table.py))
 - [ ] update docs with note for `AnyPath` that some cloudpathlib methods may only be supported on compatible Python verions
 - [ ] update HISTORY.md

pjbull:
Hi @Gilthans, would you have a chance to wrap what this PR needs? Should just be the [items above](https://github.com/drivendataorg/cloudpathlib/pull/290#issuecomment-1322713716) after rebasing onto the latest.

Gilthans:
Sure, lost track of this PR completely!

Gilthans:
> * [ ]  update docs with note for `AnyPath` that some cloudpathlib methods may only be supported on compatible Python verions
Where would you envision this appearing? In the `readme.md`? In the class docstring?

Gilthans:
Another question: obviously tests are failing for python <3.9. Is there a way to specify that this test should only run on certain python versions?

cloudpathlib/cloudpath.py
```diff
@@ -759,6 +759,9 @@ def suffix(self) -> str:
     def suffixes(self) -> List[str]:
         return self._dispatch_to_path("suffixes")
 
+    def with_stem(self, stem: str) -> Self:
+        return self._dispatch_to_path("with_stem", stem)
```
pjbull, line 763: CPython implementation is trivial, so maybe we put this in a try-except and fallback to our backport of their implementation if it is not present:
https://github.com/python/cpython/pull/19295/files#diff-fa525485738fc33d05b06c159172ff1f319c26e88d8c6bb39f7dbaae4dc4105cR859-R862

pjbull (changes requested):

cloudpathlib/cloudpath.py
pjbull: (this means that we won't need to skip the test or make a note about compatible versions)

pjbull (approved):

PR #298, opened 2022-12-14
CloudPath type hints

ringohoffman:
mainly provides improved type hints for derived classes of `CloudPath` thru the `DerivedCloudPath` `TypeVar`, e.g. now:

```python
S3Path('s3://bucket/key') / '.txt'
```

is also hinted as an `S3Path`!

general improvements to static type hinting; now `mypy cloudpathlib --strict` yields 121 errors down from 209!

made instance attributes of `CloudImplementation` into class attributes for improved hinting, necessitating a few couple line changes in the client and local path implementations

### DIFF
--- a/cloudpathlib/client.py
+++ b/cloudpathlib/client.py
@@ -61,7 +61,7 @@ class Client(abc.ABC, Generic[BoundedCloudPath]):
         self.__class__._default_client = self
 
     def CloudPath(self, cloud_path: Union[str, BoundedCloudPath]) -> BoundedCloudPath:
-        return self._cloud_meta.path_class(cloud_path=cloud_path, client=self)
+        return self._cloud_meta.path_class(cloud_path=cloud_path, client=self)  # type: ignore
 
     @abc.abstractmethod
     def _download_file(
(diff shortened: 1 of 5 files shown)


ringohoffman:
@pjbull eager to hear your thoughts on this PR

ringohoffman:
@pjbull I did want to ask about `CloudPath.copytree` and `CloudPath.copy`... are these meant to be able to copy across different subclasses of `CloudPath`? Like can I copy from an `S3Path` to a `GSPath`? It seemed like that was the intention, so I left it hinted generically.

pjbull:
Thanks @ringohoffman, this is looks pretty great—especially tackling annotating a bunch of the core cloudpath.py. I'll take a deeper look and play with it in the next couple days.


> I did want to ask about CloudPath.copytree and CloudPath.copy... are these meant to be able to copy across different subclasses of CloudPath? Like can I copy from an S3Path to a GSPath? It seemed like that was the intention, so I left it hinted generically, but I think we could take this further and return the type based on the src/dest.

Yes, it is possible to copy across providers, going through the machine running the code:
https://github.com/drivendataorg/cloudpathlib/blob/96759b0de615a0d2f434b90c23f57ee673a238fa/cloudpathlib/cloudpath.py#L776-L784

pjbull:
Thanks @ringohoffman, this seems to work great. I'll move to a local branch to make sure everything plays nice with the live test suite.

PR #300, opened 2022-12-15
Move _handle attribute up in CloudPath __init__

### LINKED ISSUE #299: self._handle is not defined in __del__
Hello! Thank you for this awesome package. I've been debugging some errors in pytest fixtures lately, specifically `ClientMismatchError`s (which are my fault, not `cloudpathlib`). Due to causing this error, the `self._handle` attribute is never set, so when `pytest` tears the test down and calls `__del__` on `CloudPath`, there is a further exception raised which pytest leaves a note about. The fix is fairly simple, it should be enough to move the definition of the `self._handle` attribute to the top of `__init__`, or else use a `try/except` block in `__del__` to catch this situation. Are you open to a patch for this?

Sample traceback:

```
python3.9/site-packages/_pytest/unraisableexception.py:78: PytestUnraisableExceptionWarning: Exception ignored in: <function CloudPath.__del__ at 0x1071b9940>

    File ".../lib/python3.9/site-packages/cloudpathlib/cloudpath.py", line 184, in __del__
      if self._handle is not None:
  AttributeError: 'S3Path' object has no attribute '_handle'

    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
```

This is also documented in pytest here: https://docs.pytest.org/en/7.1.x/how-to/failures.html#warning-about-unraisable-exceptions-and-unhandled-thread-exceptions

bryanwweber:
The _handle attribute needs to be set before any code can raise an exception. This is because if something tries to call __del__ after the exception as the process is exiting, _handle must be set to avoid an AttributeError in __del__.

Resolves #299.

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -166,6 +166,11 @@ class CloudPath(metaclass=CloudPathMeta):
     cloud_prefix: str
 
     def __init__(self, cloud_path: Union[str, "CloudPath"], client: Optional["Client"] = None):
+        
+        # handle if local file gets opened. must be set at the top of the method in case any code
+        # below raises an exception, this prevents __del__ from raising an AttributeError
+        self._handle = None
+
         self.is_valid_cloudpath(cloud_path, raise_on_error=True)
 
         # versions of the raw string that provide useful methods
@@ -189,9 +194,6 @@ class CloudPath(metaclass=CloudPathMeta):
 
         # track if local has been written to, if so it may need to be uploaded
         self._dirty = False
-
-        # handle if local file gets opened
-        self._handle = None
 
     def __del__(self):
         # make sure that file handle to local path is closed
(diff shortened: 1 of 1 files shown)


pjbull:
Thanks @bryanwweber—looks like you need to run `black` to format those lines (you can use `make format` on your local branch).

bryanwweber:
Thanks @pjbull! Fixed now 😄 That's what I get for trying to make the change on the Web interface 😝

pjbull:
Thanks @bryanwweber! I'll merge this in to a repo-local branch so we can run the live tests.

bryanwweber:
Thanks! Do I need to do anything else?

pjbull:
Nope, once tests pass in #301 I'll merge it in! It'll be in the next release, thanks!

bryanwweber:
Great thank you for the quick responses!

Issue #295, opened 2022-11-30
Enable `S3Path.mkdir`


the-matt-morris:
Just found this library, and I really love it!

I do have a use case where I'm wanting to run `mkdir()` command, but noticed that `S3Path.mkdir()` states it is not possible to create empty s3 directory:

https://github.com/drivendataorg/cloudpathlib/blob/96759b0de615a0d2f434b90c23f57ee673a238fa/cloudpathlib/s3/s3path.py#L41-L43

The method signature from the abstractmethod may not be ideal for the s3 situation (would have to play around with this), but it is possible to create an empty directory in s3.  I was using the `boto3.S3Client.put_object` method in my use case, but the proposed solution in [this comment](https://github.com/drivendataorg/cloudpathlib/issues/51#issuecomment-1245469574) looks good.

The issue with the comment referenced above does mention this problem, but I'm opening a separate issue dedicated for the `S3Path.mkdir` here because that issue has a larger scope and I believe this could be implemented quickly.  Let me know if you think this is off base, and I can close this.  Thanks!

xyXiangYu:
@the-matt-morris , this is also what I am looking for. Additionally, I would like to copy files from local directory to s3 bucket while keeping the local directory structure. When making a folder locally, we can use `path.mkdir(parents=True, exist_ok=True)`, it seems such feature is missing in S3Path.mkdir. Hope this issue can be solved soon.

xyXiangYu:
A less intuitive solution i am using now is  `CloudPath(f"{aws_s3_bucket_string}/existing_folder/subfolder_to_be_created/").touch()`. Then a `subfolder_to_be_created` directory will be created within the `existing_folder` on the AWS s3 bucket. It would be ideal to use the mkdir() method directly since it is more intuitive.

pjbull:
Thanks for the thoughts @xyXiangYu and @the-matt-morris.

I'm curious about what your use cases are. We could implement `mkdir`, and we're definitely open to doing so with demonstrated user need.

[...]

Issue #155, opened 2021-08-02
S3: Use list_objects_v2 to list objects


pjbull:
The v1 of list_objects can be the cause of some consistency problems. See this answer for context:
https://stackoverflow.com/a/67412931/1692709

We currently use `list_objects` for non recursive cases:
https://github.com/drivendataorg/cloudpathlib/blob/80f7afdf85dfb4f3ad0406944a5d3cf28c727435/cloudpathlib/s3/s3client.py#L147

We use the bucket filter in recursive cases:
https://github.com/drivendataorg/cloudpathlib/blob/80f7afdf85dfb4f3ad0406944a5d3cf28c727435/cloudpathlib/s3/s3client.py#L136

We should replace both code paths with `self.client.get_paginator('list_objects_v2')`:

Here's the boto3 docs: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.list_objects_v2

pjbull:
Closed by #302

PR #302, opened 2022-12-18
Use list_objects_v2 API for S3

### LINKED ISSUE #151: ENH: Allow .resolve() with CloudPaths
Hello,

In order to simplify the code and unify the API, would it be possible to allow the .resolve() function to CloudPaths, even if this function does nothing ?

To go from:
```python
path = AnyPath(path)
if not isinstance(path, CloudPath):
   path = path.resolve()
```

to 

```python
path = AnyPath(path).resolve()
```

Thanks a lot

### LINKED ISSUE #171: Smoke tests for non-all installations
Our CI tests currently only test the `[all]` dependency configuration. This fails to surface problems when missing optional dependencies cause runtime errors, e.g., the error fixed by #170. (Luckily, our conda-forge feedstock recipe does run tests like these and [caught that one](https://github.com/conda-forge/cloudpathlib-feedstock/pull/17/checks?check_run_id=3637922994).)

One easy way to do this is to add a job in GitHub Actions with a matrix on the different extras, and then simply run something like
```bash
python -c "from cloudpathlib.cloudpath import implementation_registry; implementation_registry['s3'].validate_completeness()"
```
like what we do for our [conda-forge feedstock](https://github.com/conda-forge/cloudpathlib-feedstock/blob/master/recipe/meta.yaml#L87).

### LINKED ISSUE #291: `exists` should work on a bucket
CloudPath currently doesn't let you easily check if a bucket itself exists. The `exists` method only works if you provide some path inside the bucket. This is a nice feature in Pathy and the analogous equivalent of `pathlib.Path` also works. It would be great to have this included in this library as well.

```python
from pathy import Pathy
from pathlib import Path

# Doesn't work
CloudPath("s3://some-bucket/").exists()
# ParamValidationError: Parameter validation failed:
# Invalid length for parameter Key, value: 0, valid min length: 1

Pathy("s3://some-bucket/").exists() # True

# Since a bucket is the equivalent of a drive, a comparison to pathlib.Path 
# for checking if a drive exists (I'm using Windows)
 Path("C:").exists() # True
```

pjbull:
This is primarily a fix for #155, but it should simplify and stabilize listing dirs from S3 in a way that makes a number of the other directory handling tasks easier (like #295, #181, #176, and #51).

One of the concerns is that this change may affect performance since we change both the recursive and non-recursive apis, so this PR includes an initial perf test suite for listing directories. See results below (TLDR: negligible change if any).

Bonus fixes:
 - Fix #171 (approved code would have introduced an error here, so worth it to fix)
 - Perf test suite
 - test-debug command stops debugger at failing tests from previous run
 - Check existence of buckets/containers when permitted by permissions
 - Add `pytest-xdist` to requirements, which will run the test suite in parallel and results in decent speed up. Comparison of latest `master` run and this branch shaves off ~10 minutes (makes sense since gh action machines have 2 cores):
image

 
Closes #151
Closes #171 
Closes #291

### Before changes

```
                                 Performance suite results: (2022-12-17T16:11:32.734517)
┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ Test Name    ┃ Config Name                ┃ Iterations ┃           Mean ┃              Std ┃            Max ┃ N Items ┃
┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│ List Folders │ List shallow recursive     │         10 │ 0:00:01.425719 │ ± 0:00:00.172902 │ 0:00:01.788437 │   5,500 │
│ List Folders │ List shallow non-recursive │         10 │ 0:00:01.630402 │ ± 0:00:00.187300 │ 0:00:01.882856 │   5,500 │
│ List Folders │ List normal recursive      │         10 │ 0:00:02.168442 │ ± 0:00:00.039756 │ 0:00:02.219833 │   7,877 │
│ List Folders │ List normal non-recursive  │         10 │ 0:00:00.039067 │ ± 0:00:00.010252 │ 0:00:00.067796 │     113 │
│ List Folders │ List deep recursive        │         10 │ 0:00:03.237524 │ ± 0:00:00.060835 │ 0:00:03.317934 │   7,955 │
│ List Folders │ List deep non-recursive    │         10 │ 0:00:00.030883 │ ± 0:00:00.001670 │ 0:00:00.033317 │      31 │
└──────────────┴────────────────────────────┴────────────┴────────────────┴──────────────────┴────────────────┴─────────┘
```

### With new changes

```
                                 Performance suite results: (2022-12-17T17:11:50.937390)
┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ Test Name    ┃ Config Name                ┃ Iterations ┃           Mean ┃              Std ┃            Max ┃ N Items ┃
┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│ List Folders │ List shallow recursive     │         10 │ 0:00:01.293559 │ ± 0:00:00.027168 │ 0:00:01.336273 │   5,500 │
│ List Folders │ List shallow non-recursive │         10 │ 0:00:01.289645 │ ± 0:00:00.037398 │ 0:00:01.348425 │   5,500 │
│ List Folders │ List normal recursive      │         10 │ 0:00:01.874748 │ ± 0:00:00.079183 │ 0:00:02.006050 │   7,877 │
│ List Folders │ List normal non-recursive  │         10 │ 0:00:00.033232 │ ± 0:00:00.004852 │ 0:00:00.045547 │     113 │
│ List Folders │ List deep recursive        │         10 │ 0:00:03.193576 │ ± 0:00:00.180428 │ 0:00:03.455747 │   7,955 │
│ List Folders │ List deep non-recursive    │         10 │ 0:00:00.030362 │ ± 0:00:00.001615 │ 0:00:00.032398 │      31 │
└──────────────┴────────────────────────────┴────────────┴────────────────┴──────────────────┴────────────────┴─────────┘
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,3 +82,6 @@ test-debug:
 
 test-live-cloud:  ## run tests on live cloud backends
 	USE_LIVE_CLOUD=1 python -m pytest -vv
+
+perf:
+	python tests/performance/cli.py s3 --save-csv=perf-results.csv
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ docs: clean-docs docs-setup ## build the static version of the docs
 docs-serve: clean-docs docs-setup ## serve documentation to livereload while you work
 	cd docs && mkdocs serve
 
-format:
+format:  ## run blank for format codebase
 	black cloudpathlib tests docs
 
 help:
@@ -60,7 +60,7 @@ help:
 install: clean ## install the package to the active Python's site-packages
 	python setup.py install
 
-lint: ## check style with flake8
+lint: ## check style with black, flake8, and mypy
 	black --check cloudpathlib tests docs
 	flake8 cloudpathlib tests docs
 	mypy cloudpathlib
@@ -71,17 +71,17 @@ release: dist ## package and upload a release
 release-test: dist
 	twine upload --repository pypitest dist/*
 
-reqs:
+reqs:  ## install development requirements
 	pip install -U -r requirements-dev.txt
 
 test: ## run tests with mocked cloud SDKs
 	python -m pytest -vv
 
-test-debug:
+test-debug:  ## rerun tests that failed in last run and stop with pdb at failures
 	python -m pytest -n=0 -vv --lf --pdb
 
 test-live-cloud:  ## run tests on live cloud backends
 	USE_LIVE_CLOUD=1 python -m pytest -vv
 
-perf:
+perf:  ## run performance measurement suite for s3 and save results to perf-results.csv
 	python tests/performance/cli.py s3 --save-csv=perf-results.csv
--- a/Makefile
+++ b/Makefile
@@ -77,5 +77,8 @@ reqs:
 test: ## run tests with mocked cloud SDKs
 	python -m pytest -vv
 
+test-debug:
+	python -m pytest -n=0 -vv --lf --pdb
+
 test-live-cloud:  ## run tests on live cloud backends
 	USE_LIVE_CLOUD=1 python -m pytest -vv
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ docs: clean-docs docs-setup ## build the static version of the docs
 docs-serve: clean-docs docs-setup ## serve documentation to livereload while you work
 	cd docs && mkdocs serve
 
-format:  ## run blank for format codebase
+format:  ## run black to format codebase
 	black cloudpathlib tests docs
 
 help:
--- a/tests/mock_clients/mock_azureblob.py
+++ b/tests/mock_clients/mock_azureblob.py
@@ -12,6 +12,9 @@ from .utils import delete_empty_parents_up_to_root
 TEST_ASSETS = Path(__file__).parent.parent / "assets"
 
 
+DEFAULT_CONTAINER_NAME = "container"
+
+
 def mocked_client_class_factory(test_dir: str):
     class MockBlobServiceClient:
         def __init__(self, *args, **kwargs):
@@ -111,7 +114,7 @@ class MockContainerClient:
         self.container_name = container_name
 
     def exists(self):
-        if self.container_name == "container":  # name used by passing tests
+        if self.container_name == DEFAULT_CONTAINER_NAME:  # name used by passing tests
             return True
         else:
             return False
--- a/tests/mock_clients/mock_azureblob.py
+++ b/tests/mock_clients/mock_azureblob.py
@@ -33,7 +33,7 @@ def mocked_client_class_factory(test_dir: str):
             return MockBlobClient(self.tmp_path, blob, service_client=self)
 
         def get_container_client(self, container):
-            return MockContainerClient(self.tmp_path)
+            return MockContainerClient(self.tmp_path, container_name=container)
 
     return MockBlobServiceClient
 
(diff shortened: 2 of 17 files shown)


Makefile
```diff
@@ -51,7 +51,7 @@ docs: clean-docs docs-setup ## build the static version of the docs
 docs-serve: clean-docs docs-setup ## serve documentation to livereload while you work
 	cd docs && mkdocs serve
 
-format:
+format:  ## run blank for format codebase
```
jayqi, line 54:
```suggestion
format:  ## run black to format codebase
```

tests/mock_clients/mock_azureblob.py
```diff
@@ -106,8 +106,15 @@ def content_as_bytes(self):
 
 
 class MockContainerClient:
-    def __init__(self, root):
+    def __init__(self, root, container_name):
         self.root = root
+        self.container_name = container_name
+
+    def exists(self):
+        if self.container_name == "container":  # name used by passing tests
```
jayqi, line 114: Wondering if this would be more clear if we had this as a global constant, and had conftest instead import it from here. Right now these are just hardcoded values in two different places.

tests/performance/cli.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+from runner import main, normalize_results, results_to_csv
+
+# make loguru and tqdm play nicely together
+logger.remove()
+logger.add(lambda msg: tqdm.write(msg, end=""), colorize=True)
+
+# get environement variables
+load_dotenv(find_dotenv())
```
jayqi, line 24: I personally dislike automatic `.env` loading like this, and prefer having it be more explicit / closer to the user. For example, have the make command use the dotenv CLI with `dotenv -f .env run python tests/performance/cli.py s3 --save-csv=perf-results.csv`. Not a position I hold strongly.

jayqi (approved):
Looks good. One comment typo, and a few minor comments.

tests/performance/cli.py
pjbull: Yeah, agreed this should be as close to the user as possible (which is why we have it just at the `cli.py` entrypoint and `conftest`). The library code does not explicitly do anything with `.env`. I have a slight preference for loading this at the entrypoint since I don't always use these commands through the makefile, and I expect my `.env` to get loaded (and I also want parsimonious cli commands).

tests/mock_clients/mock_azureblob.py
pjbull: Good call, did this for all the defaults.

PR #304, opened 2022-12-20
Speed up glob implementation

### LINKED ISSUE #274: glob and rglob can be slow for large directories or lots of files

pjbull:
I believe that globs were slow because the internals of how pathlib calls our functions made us loop over all of the files in the directory every time it was checking children, which happened for any nested file.

Instead of passing the whole file tree to every selectable instance and letting that object filter it down on demand, this implementation builds the entire file tree as nested dicts and then lets `scandir` walk those nested dicts. Each selectable now knows only about its direct parents/children.

Closes #274

# Perf results

The list and glob scenarios listed below are executed against the same file trees, so this lets us more or less compare raw `_list_dir` calls to the performance of glob. (Note: N: items is different because the glob test filters out folders).

## Before this change

Super slow........

```
                                  Performance suite results: (2022-12-20T01:29:29.267986)                                  
┏━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ Test Name      ┃ Config Name                ┃ Iterations ┃           Mean ┃              Std ┃            Max ┃ N Items ┃
┡━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│ List Folders   │ List shallow recursive     │          2 │ 0:00:01.256078 │ ± 0:00:00.037748 │ 0:00:01.282770 │   5,500 │
│ List Folders   │ List shallow non-recursive │          2 │ 0:00:01.261018 │ ± 0:00:00.003631 │ 0:00:01.263586 │   5,500 │
│ List Folders   │ List normal recursive      │          2 │ 0:00:01.778471 │ ± 0:00:00.019042 │ 0:00:01.791936 │   7,877 │
│ List Folders   │ List normal non-recursive  │          2 │ 0:00:00.035925 │ ± 0:00:00.000327 │ 0:00:00.036156 │     113 │
│ List Folders   │ List deep recursive        │          2 │ 0:00:03.105114 │ ± 0:00:00.059515 │ 0:00:03.147197 │   7,955 │
│ List Folders   │ List deep non-recursive    │          2 │ 0:00:00.040619 │ ± 0:00:00.010220 │ 0:00:00.047846 │      31 │
│ Glob scenarios │ Glob shallow recursive     │          2 │ 0:06:58.346311 │ ± 0:00:00.956693 │ 0:06:59.022795 │   5,500 │
│ Glob scenarios │ Glob shallow non-recursive │          2 │ 0:04:39.675058 │ ± 0:00:00.115285 │ 0:04:39.756577 │   5,500 │
│ Glob scenarios │ Glob normal recursive      │          2 │ 0:00:54.873498 │ ± 0:00:00.014340 │ 0:00:54.883638 │   7,272 │
│ Glob scenarios │ Glob normal non-recursive  │          2 │ 0:00:06.609309 │ ± 0:00:00.005283 │ 0:00:06.613045 │      12 │
│ Glob scenarios │ Glob deep recursive        │          2 │ 0:04:09.120405 │ ± 0:00:01.541716 │ 0:04:10.210563 │   7,650 │
│ Glob scenarios │ Glob deep non-recursive    │          2 │ 0:00:05.651228 │ ± 0:00:00.232796 │ 0:00:05.815840 │      25 │
└────────────────┴────────────────────────────┴────────────┴────────────────┴──────────────────┴────────────────┴─────────┘
```

## After this change

Very little overhead when doing actual comparisons and filtering (~10%) beyond the vanilla `_list_dir` calls!

```
                                  Performance suite results: (2022-12-20T09:26:30.055262)                                  
┏━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ Test Name      ┃ Config Name                ┃ Iterations ┃           Mean ┃              Std ┃            Max ┃ N Items ┃
┡━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│ List Folders   │ List shallow recursive     │         10 │ 0:00:01.213074 │ ± 0:00:00.047582 │ 0:00:01.307935 │   5,500 │
│ List Folders   │ List shallow non-recursive │         10 │ 0:00:01.235682 │ ± 0:00:00.020873 │ 0:00:01.276783 │   5,500 │
│ List Folders   │ List normal recursive      │         10 │ 0:00:01.764192 │ ± 0:00:00.059765 │ 0:00:01.825311 │   7,877 │
│ List Folders   │ List normal non-recursive  │         10 │ 0:00:00.036764 │ ± 0:00:00.004089 │ 0:00:00.047984 │     113 │
│ List Folders   │ List deep recursive        │         10 │ 0:00:03.052238 │ ± 0:00:00.457676 │ 0:00:04.332024 │   7,955 │
│ List Folders   │ List deep non-recursive    │         10 │ 0:00:00.032839 │ ± 0:00:00.002672 │ 0:00:00.038686 │      31 │
│ Glob scenarios │ Glob shallow recursive     │         10 │ 0:00:01.484482 │ ± 0:00:00.065573 │ 0:00:01.561880 │   5,500 │
│ Glob scenarios │ Glob shallow non-recursive │         10 │ 0:00:01.481584 │ ± 0:00:00.029798 │ 0:00:01.523332 │   5,500 │
│ Glob scenarios │ Glob normal recursive      │         10 │ 0:00:02.017116 │ ± 0:00:00.055220 │ 0:00:02.115873 │   7,272 │
│ Glob scenarios │ Glob normal non-recursive  │         10 │ 0:00:00.037116 │ ± 0:00:00.003030 │ 0:00:00.043329 │      12 │
│ Glob scenarios │ Glob deep recursive        │         10 │ 0:00:03.404524 │ ± 0:00:00.244392 │ 0:00:04.028038 │   7,650 │
│ Glob scenarios │ Glob deep non-recursive    │         10 │ 0:00:00.033715 │ ± 0:00:00.002128 │ 0:00:00.036057 │      25 │
└────────────────┴────────────────────────────┴────────────┴────────────────┴──────────────────┴────────────────┴─────────┘
```

Also, CI timings are fickle, but we may have knocked a few minutes off the test suite when compared to `master`:
image

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -363,29 +363,28 @@ class CloudPath(metaclass=CloudPathMeta):
         # build a tree structure for all files out of default dicts
         Tree: Callable = lambda: defaultdict(Tree)
 
+        def _build_tree(trunk, branch, nodes, is_dir):
+            """Utility to build a tree from nested defaultdicts with a generator
+            of nodes (parts) of a path."""
+            next_branch = next(nodes, None)
+
+            if next_branch is None:
+                trunk[branch] = Tree() if is_dir else None  # leaf node
+
+            else:
+                _build_tree(trunk[branch], next_branch, nodes, is_dir)
+
         file_tree = Tree()
 
         for f, is_dir in self.client._list_dir(self, recursive=recursive):
             parts = str(f.relative_to(self)).split("/")
 
-            if len(parts) == 1:
-                # skip self
-                if parts[0] == ".":
-                    continue
+            # skip self
+            if len(parts) == 1 and parts[0] == ".":
+                continue
 
-                file_tree[parts[0]] = Tree() if is_dir else None
-
-            else:
-                # put top-level folder in file tree
-                node = file_tree[parts[0]]
-
-                # add remaining folders
-                for part in parts[1:-1]:
-                    node = node[part]
-
-                # add last item pointing at empty Tree if it is a folder
-                # else pointing at nothing if it is a file
-                node[parts[-1]] = Tree() if is_dir else None
+            nodes = (p for p in parts)
+            _build_tree(file_tree, next(nodes, None), nodes, is_dir)
 
         file_tree = dict(file_tree)  # freeze as normal dict before passing in
 
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -357,44 +357,67 @@ class CloudPath(metaclass=CloudPathMeta):
         if pattern.startswith(self.cloud_prefix) or pattern.startswith("/"):
             raise CloudPathNotImplementedError("Non-relative patterns are unsupported")
 
-    def _glob(self: DerivedCloudPath, selector) -> Generator[DerivedCloudPath, None, None]:
+    def _glob(
+        self: DerivedCloudPath, selector, recursive: bool
+    ) -> Generator[DerivedCloudPath, None, None]:
+        # build a tree structure for all files out of default dicts
+        Tree: Callable = lambda: defaultdict(Tree)
+
+        file_tree = Tree()
+
+        for f, is_dir in self.client._list_dir(self, recursive=recursive):
+            parts = str(f.relative_to(self)).split("/")
+
+            if len(parts) == 1:
+                # skip self
+                if parts[0] == ".":
+                    continue
+
+                file_tree[parts[0]] = Tree() if is_dir else None
+
+            else:
+                # put top-level folder in file tree
+                node = file_tree[parts[0]]
+
+                # add remaining folders
+                for part in parts[1:-1]:
+                    node = node[part]
+
+                # add last item pointing at empty Tree if it is a folder
+                # else pointing at nothing if it is a file
+                node[parts[-1]] = Tree() if is_dir else None
+
+        file_tree = dict(file_tree)  # freeze as normal dict before passing in
+
         root = _CloudPathSelectable(
-            PurePosixPath(self._no_prefix_no_drive),
-            {
-                PurePosixPath(c._no_prefix_no_drive): is_dir
-                for c, is_dir in self.client._list_dir(self, recursive=True)
-            },
-            is_dir=True,
-            exists=True,
+            self.name,
+            [p.name for p in self.parents[:-1]],  # all parents except bucket/container
+            file_tree,
         )
 
         for p in selector.select_from(root):
-            yield self.client.CloudPath(f"{self.cloud_prefix}{self.drive}{p}")
+            yield self.client.CloudPath(f"{self.cloud_prefix}{self.drive}/{p}")
 
     def glob(self: DerivedCloudPath, pattern: str) -> Generator[DerivedCloudPath, None, None]:
-        if pattern == "*":
-            yield from (path for path, is_file in self.client._list_dir(self, recursive=False))
-            return
-        if pattern == "**/*":
-            yield from (path for path, is_file in self.client._list_dir(self, recursive=True))
-            return
         self._glob_checks(pattern)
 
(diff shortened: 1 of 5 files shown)


cloudpathlib/cloudpath.py
```diff
@@ -357,44 +357,66 @@ def _glob_checks(self, pattern: str) -> None:
...
         pattern_parts = PurePosixPath(pattern).parts
         selector = _make_selector(tuple(pattern_parts), _posix_flavour)
 
-        yield from self._glob(selector)
+        yield from self._glob(
+            selector,
+            "/" in pattern
+            or "**"
+            in pattern,  # recursive listing needed if explicit ** or any sub folder in pattern
+        )
```
jayqi, line 411: Some food for thought: Wondering if this is still searching a larger part of the file tree than needed.

If you have `Cloudpath("s3://bucket/depth1").glob("depth2/depth3/*.txt")`, it seems like this will recursively search `s3://bucket/depth1` (i.e., `self`). But what we really care about is a non-recursive listing of `s3://bucket/depth1/depth2/depth3`. If any of the parent directories have sibling directories with a lot of files, then I think this ends up being more expensive than necessary? Maybe a further optimization to, rather than doing `self._glob`, construct a `CloudPath` from all pattern parts up until one with a `*` in it? 

Maybe this is an unnecessary optimization. Or maybe there's something about how pathlib's glob works that makes this not make sense—I don't really remember the details about what was wrong with our old glob.

jayqi (approved):
Very cool! One partially-baked thought but this looks like a big improvement and tests passing seems convincing.

cloudpathlib/cloudpath.py
pjbull: This is a good point, and I started to think about optimizations, but got worried that I'd miss some idiosyncrasy about glob logic or re-implement too much of it naïvely.

To me, I think the current overhead seems pretty acceptable, but I'll file an issue from this to see if anyone has glob perf complaints after this and we want to optimize more.

pjbull:
Thanks!

PR #307, opened 2022-12-29
Updates to S3Client internals to pass extras and more

### LINKED ISSUE #48: Ability to list buckets / containers that the auth'd account has access to
Right now you need to know your bucket or container ahead of time, but most services provide a way to list the buckets or containers you can access.

Could be a separate cloud only function, could be on `CloudPath("s3://").iterdir()`

### LINKED ISSUE #180: RequestPayer option for S3
Hi,

Thank you for the efforts on this package - I appreciate your work! I've got feature proposal because I stumbled about this today and dont think it is implemented.

AWS sometimes has the option of "Requester pays" e.g. on access of public datasets. You'll have to pass the credentials and additionally the keyword args `RequestPayer='requester'` in boto3.

Documentation: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ObjectsinRequesterPaysBuckets.html

As far as I see this option is not included yet in the codebase of cloudpathlib.
It should be integrated to the bucket.objects.filter() function of boto3, e.g.

```py
bucket.objects.filter(Prefix=prefix, RequestPayer='requester')
```

What do you think?

Best regards,
Johannes

### LINKED ISSUE #212: Overly specific asserts on S3Client._remove
Thanks a bunch for this library, I've been trying to make use of it with an s3-emulating service that is not 100% bit-for-bit compliant and I've run into issues in a couple places.
Specifically I've noticed that in S3Client._remove, you assert for response==204, the clone-service I'm using responds with a 200 when you successfully delete instead.
I'd love to submit a PR to fix this but I need some input.
I think it should be acceptable to check if the response is a 2xx code and by doing so you're more flexible. In addition, asserts outside of unit testing are considered hazardous, instead this should raise an exception of some kind.
Do you have a preferred exception to be raised here instead? or should it be left to boto to raise the client error?

### LINKED ISSUE #254: "ExtraArgs" option is not available while uploading a file/folder to S3 via CloudPath
If users want to add User defined metadata tags while uploading to S3 they should be able to do so, but currently can't find a way for this using CloudPath function upload_from(). It would be great to have this implementation and I can contribute to enabling this feature if needed.

### LINKED ISSUE #271: Update docs: Use no_sign_request to fix "NoCredentials" Error when trying to access a public S3 bucket
I've outlined this issue in more depth here: https://stackoverflow.com/q/73708204/15739035

But essentially just using this simple code:
```
path = r"C:\some\path\to\folder"
url = "s3://some-example-bucket/folder/"

cloud_path = CloudPath(url)
cloud_path.download_to(path)
```

I'm unable to download a folder from this S3 bucket: `s3://cozy-auto-texture-sd-repo/stable-diffusion-v1-4/`

Now this bucket needs to be public so I can distribute a repository to my users, and anyone needs to be able to access it. For some reason whenever I run the above code with that s3:// link, `cloudpathlib` gives the following error:
```
  File "E:\Cozy-Auto-Texture-Files\venv\lib\site-packages\botocore\auth.py", line 418, in add_auth
    raise NoCredentialsError()
botocore.exceptions.NoCredentialsError: Unable to locate credentials
```

pjbull:
Updates a number of issues with S3Client internals, primarily the ability to pass `extra_args` that may be necessary to a user. This is primarily to address #254 and #180, which require passing extras around at the client level.

The current implementation assumes that all of the upload/download/listing operations will always use the same extras at the level of a client. This is reasonable for most of the extra args except some of the upload ones (Metadata, ContentType, etc.). For the ContentType and ContentEncoding, we already provide a separate means of setting these per-file.

Additionally, we add documentation for RequestPayer explicitly and then ExtraArgs generally.

Bonus:
 - #48 listing buckets across providers (and tests)
 - #271 Adds an example of a public bucket and caveat that listing permissions may not be available
 - Move the `super().__init__` calls in the azure objects so that credential errors don't result in additional warnings about missing properties
 - Simplify `_s3_file_query` logic now that API changes make it simpler to tell if something is a file or dir immediately
 - #212 Fix using `assert` in production code and status code checks that were too specific

Closes #48 
Closes #254 
Closes #180 
Closes #271
Closes #212

### DIFF
--- a/cloudpathlib/azure/azblobclient.py
+++ b/cloudpathlib/azure/azblobclient.py
@@ -73,6 +73,8 @@ class AzureBlobClient(Client):
             content_type_method (Optional[Callable]): Function to call to guess media type (mimetype) when
                 writing a file to the cloud. Defaults to `mimetypes.guess_type`. Must return a tuple (content type, content encoding).
         """
+        super().__init__(local_cache_dir=local_cache_dir, content_type_method=content_type_method)
+
         if connection_string is None:
             connection_string = os.getenv("AZURE_STORAGE_CONNECTION_STRING", None)
 
@@ -89,8 +91,6 @@ class AzureBlobClient(Client):
                 "AzureBlobClient does not support anonymous instantiation. "
                 "Credentials are required; see docs for options."
             )
-
-        super().__init__(local_cache_dir=local_cache_dir, content_type_method=content_type_method)
 
     def _get_metadata(self, cloud_path: AzureBlobPath) -> Union["BlobProperties", Dict[str, Any]]:
         blob = self.service_client.get_blob_client(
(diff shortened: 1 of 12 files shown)


jayqi (approved):

pjbull:
Thanks @jayqi !

Issue #315, opened 2023-01-24
Some libraries generate confusing errors when opening CloudPath objects


pjbull:
The main problem is the libs based on opening files with `fsspec` won't work properly with `CloudPaths`.
They will call `fspath` to resolve it (instead of str()) and therefore the file called will be cached (this is not exaclty what we want, and this will be done silently)

For single files like `tif` this will be not optimal but will work (your example)
But for nested files likes [VRT](https://gdal.org/drivers/raster/vrt.html), this will crash as only the header is cached and not the relative files. This will break the paths inside the VRT.
Note that GDAL (and therefore rasterio) can read VRT stored on the cloud, and this is the expected behavior.

A little example:
```python

vrt = AnyPath("https://s3.unistra.fr/merge_32-31.vrt")
rasterio.open(str(vrt)) # will work
rasterio.open(vrt) # will fail for the wrong reason, saying that the files linked inside the VRT don't exist
```

_Originally posted by @remi-braun in https://github.com/drivendataorg/cloudpathlib/issues/96#issuecomment-1402068539_

pjbull:
There aren't great options here. This is what I see. Open to other options if people have them. Maybe some combination of these are good mitigations.


## Stop being `os.PathLike`

A lot of these libraries do something like:

```python
def open(f):
    is instance(f, os.PathLike):
        str_path = os.fspath(f)
    ...
```

They then have special-case logic to handle str_path and work with the local filesystem path. This only works with `CloudPath`s in read scenarios for single files. Other scenarios (writing, multi-file formats, directory walking) are all broken with code that works like this.

When the code eventually goes looking on the filesystem relative to `str_path` it errors in potentially confusing ways.

If we are no longer `os.PathLike` then code will likely error earlier with clearer error messages.

Pros:
 - Possibly clearer error messages

Cons:
 - Breaks single-file read scenarios, which are _very_ common. 


## Cache entire directory on `fspath` call

We could do something like the following (and make this an option)

```python
def __fspath__(self):
    if self.is_dir():
        self.download_to(self._local)  # cache entire directory recursively on `fspath` call      
```

Pros: 
 - No missing files when relative ones are referred to.
 - Could be optional based on setting/envvar

Cons:
 - Likely a lot of wasted network traffic and disk space usage (depending on what is in the directory)


## Warn on `__fspath__` being called

We could just always warn about writing/multi-file whenever `__fspath__` is called.

```python
def __fspath__(self):
[...]

Issue #73, opened 2020-09-05
Add ability to detect if cache is dirty and should be uploaded even if file was not opened by cloudpathlib


pjbull:
We all want to live in a world where every Python library hands PathLike objects. This is not that world.

Many libraries need a path to a local file—especially as a string—in order to read that file. We should expose a supported and documented way to get a path to the version in the cache.

This came up for me in working with `pyvips` where I had to do something like this:
```python
s3p = S3Path("s3://...")

# actually do the download
s3p._refresh_cache()

pyvips.read_image(str(s3p._local))
```

This is not ideal. One option (as noted here: https://github.com/drivendataorg/cloudpathlib/pull/72 ) is to override `__fspath__` to do the caching and return the local path. Then something like this would work:

```python
pyvips.read_image(str(Path(S3Path("s3://...")))
```

Another option (or in addition) is to add a property like `read_only_local_path_string` (🤣 at name)

```python
pyvips.read_image(S3Path("s3://...").read_only_local_path_string)
```

The big caveat is that we replace the `.close` method on the buffer if you open for write through CloudPath so that we know you intend to change the file:
https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/cloudpath.py#L322-L339

Anything you do to the local path is pretty much read-only since we won't automatically upload. (At least, not without a big change—for example, overriding `__del__` on the CloudPath to do the upload if local is newer, or tracking all of the files DL'd by a Client on the Client and having it check modified times on those and upload. In general I am a little worried about automatically assuming a user wants changed files to be uploaded to overwrite things on the cloud.... )

chapmanjacobd:
Is this already supported well? I think this issue can be closed

https://cloudpathlib.drivendata.org/caching/#Handling-conflicts

If people don't want to use the force_overwrite_to_cloud kwarg in `.Open` then they could do custom programming using the errors:

```py
s3p = S3Path("s3://...")

try:
    pyvips.read_image(s3p.fspath)

except OverwriteNewerCloudError:
    s3p._refresh_cache()
    pyvips.read_image(s3p.fspath)
```

pjbull:
This issue is somewhat mitigated by the work that has gone into #140.

That said, it tracks the generic case of:
 - Use `CloudPath` so that file is downloaded to local cache
 - Local cache gets edited in some way we don't know about--e.g., shell out to a separate process that edits the file.
 - User may expect changes to be re-uploaded to the cloud, but they are not.

[...]

PR #322, opened 2023-02-16
WIP: Patch open, os and os.path builtins

pjbull:
Right now, libs that use `open` directly, `fspath` or just assume paths are strings and can be manipulated don't work with `CloudPath` objects. (Related issues: #315, #73, #128)

Totally WIP and not sure that we want to introduce and support this, but it may increase compatibility significantly to have the option at least for users to patch the builtin functions.

The biggest limitations are:
 - Libs that do any string manipulation manually still will not work (e.g. `path = dir + os.sep + file`)
 - Libs that pass local paths off to compiled code to do IO will still not work
 - There are probably lots of other ways that other code can do funky things with paths, so even these patches may not make things work

Other potential todos for this PR:
 - [x] Patch the standard lib `glob` module as well
 - [x] Add integration test suite that shows the patches working with libs that did not work previously (`pandas` has all kinds of formats, `rasterio` may be interesting, `xarray` has come up, `Pillow` may be a common use case) 
 - [x] Add env var for patching `os` functions
 - [x] override `os.fspath` to allow returning `CloudPath`
 - [x] Document these methods
 - [x] Add full test suite (can we leverage what CPython does for these functions?)
 - [x] Document that in Jupyter notebooks you have to do `open = patch_open()` since `open` exists as a `local` in the notebook namespace
 - [x] Remove testing script

Curious for your thoughts @jayqi !

-------------------

## Experimental instructions

 - `pip install git+https://github.com/drivendataorg/cloudpathlib@patch-builtins`
 - Set the environment variable `CLOUDPATHLIB_PATCH_OPEN=True` before any import call to `cloudpathlib`
 - Experiment with your code that uses `open` or third-party libraries that use open.
 - Report the results in #128


cloudpathlib/patches.py
```diff
@@ -0,0 +1,210 @@
...
+    return _patched_version
+
+
+def patch_open():
+    patched = _patch_factory(
+        builtins.open,
+        CloudPath.open,
+    )
+    builtins.open = patched
+    CloudPath.__fspath__ = lambda x: x  # turn off `fspath`
```
pjbull, line 35: So, the idea here is that even if code uses `fspath` to try to get the local path, we give back the `CloudPath` instead so it works downstream. This may break code that assumes `fspath` returns a string, which is what it is supposed to do.

pjbull:
@jayqi ok, I think that this is finally ready to ship if test suite passes! Let me know if you want to do a review—it's pretty beefy so could do synchronous if easier.

Makefile
```diff
@@ -81,7 +81,7 @@ test: ## run tests with mocked cloud SDKs
 	python -m pytest -vv
 
 test-debug:  ## rerun tests that failed in last run and stop with pdb at failures
-	python -m pytest -n=0 -vv --lf --pdb
+	python -m pytest -n=0 -vv --lf --pdb --capture=no
```
pjbull, line 84: Remove?

tests/test_patching.py
```diff
@@ -0,0 +1,661 @@
...
+    finally:
+        # Clean up local file
+        os.unlink(local_path)
+
+
+def test_patch_glob_basic(rig):
+    """Test that glob functions are callable when patched."""
+    with patch_glob():
+        assert callable(glob.glob)
+        assert callable(glob.iglob)
```
pjbull, line 303: We can remove

HISTORY.md
```diff
@@ -4,6 +4,7 @@
 
 - Fixed issue with GS credentials, using default auth enables a wider set of authentication methods in GS (Issue [#390](https://github.com/drivendataorg/cloudpathlib/issues/390), PR [#514](https://github.com/drivendataorg/cloudpathlib/pull/514), thanks @ljyanesm) 
 - Added support for http(s) urls with `HttpClient`, `HttpPath`, `HttpsClient`, and `HttpsPath`. (Issue [#455](https://github.com/drivendataorg/cloudpathlib/issues/455), PR [#468](https://github.com/drivendataorg/cloudpathlib/pull/468))
+- Added support for patching the builtins `open`, `os`, `os.path`, and `glob` to work with `CloudPath` objects. (Issue [#128](https://github.com/drivendataorg/cloudpathlib/issues/128), PR [#322](https://github.com/drivendataorg/cloudpathlib/pull/322))
```
pjbull, line 7: - [ ] Signpost as experimental (here and in docs)
- [ ] Add off by default and needs to be turned on, see docs.
- [ ] Add note on supporting `parts` and that `CloudPath("s3://...", my_client)` will break (relies on kwarg order)

pjbull (commented):
A few tweaks before shipping

PR #327, opened 2023-04-16
Use Self type from PEP 673

jayqi:
This PR replaces the custom type variable `DerivedCloudPath` with the special type `Self` added in Python 3.11 (adds `typing_extensions` backport dependency for older Python versions). This works in a functionally identical way as what we had before, as explained in the [official docs](https://docs.python.org/3/library/typing.html#typing.Self), but is a part of the standard library and much more readable. 

For more context, see [PEP 673](https://peps.python.org/pep-0673/).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 ## v0.14.0 (Unreleased)
 
 - Changed to pyproject.toml-based build.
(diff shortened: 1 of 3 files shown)


pjbull (approved):

pjbull:
Nice!

Issue #328, opened 2023-05-08
Mypy error creating a `CloudPath` instance


connorbrinton:
## Description

Currently, type-checking a very simple script that uses `cloudpathlib` with Mypy fails with an error.

Script:
```python
from cloudpathlib import CloudPath

path = CloudPath("s3://abc/xyz")
```

Mypy error:
```python
test.py:3: error: Cannot instantiate abstract class "CloudPath" with abstract attributes "drive", "is_dir", "is_file", "mkdir" and "touch"  [abstract]
```

## Cause

I believe this is a result of the fact that [Mypy doesn't yet understand the `__call__` method on metaclasses](https://github.com/python/mypy/issues/14122).

### Specific Subclass

One way to avoid this error is for end users to directly use implementation-specific `CloudPath` subclasses. The following script produces no Mypy errors:
```python
from cloudpathlib import S3Path

path = S3Path("s3://abc/xyz")
```

This slightly limits portability of end user scripts, but is one possible way to avoid the error.

### Ignore Directive

The error can also be ignored with a `# type: ignore[abstract]` directive:
```python
from cloudpathlib import CloudPath

path = CloudPath("s3://abc/xyz")  # type: ignore[abstract]
```

### New Method

Lastly, some changes could be made to how `CloudPath` instances are constructed to work around this issue. However, this would introduce a new way of instantiating `CloudPath` instances, which could be confusing.

One option would be to add a `CloudPath.from_url` class method, which would just `return CloudPath(url)`. This way, the Mypy error could be ignored in the `cloudpathlib` codebase rather than user code.

Another option would be to implement cloud provider-specific dispatching in the new `CloudPath.from_url` class method, then have `CloudPathMeta.__call__` delegate to `CloudPath.from_url` for its implementation. This wouldn't require any new ignore directives in the `cloudpathlib` codebase.

In any case, I didn't find any issues describing this behavior elsewhere, so hopefully this is helpful to others who run into this same issue in the future 🙂 

Thanks for maintaining cloudpathlib! 😄

pjbull:
Thanks for the detailed report @connorbrinton. I'd be curious if #330 helps with this issue or not.

Hopefully there's an upstream `mypy` fix for `__call__` because the ergonomics of the existing API are nice as is....

jmarshall:
I can confirm that the diagnostic is still present with cloudpathlib-0.15.1, which includes #330. (This is with mypy 1.5.1.)

[...]

PR #330, opened 2023-05-11
Fix CloudPathMeta.__call__ return type

ringohoffman:
Since `pyright==1.1.307`, instances of the subclasses of `CloudPath` stopped being recognized as instances of their own class; a possible fix is to type hint `CloudPathMeta.__call__`, but it creates a host of `mypy` errors (5)... 

Here is the typing bug starting in the newest version of pyright and the fix with this proposed change:

```python
import cloudpathlib


# before
x = cloudpathlib.S3Path("s3://bucket/key")
# reveal_type(x)  # Pylance: Type of "x" is "CloudPath | Unknown | Self@__new__"
y = cloudpathlib.CloudPath(x)
# reveal_type(y)  # Pylance: Type of "y" is "CloudPath | Unknown | Self@__new__"

# after
x = cloudpathlib.S3Path("s3://bucket/key")
reveal_type(x)  # Pylance: Type of "x" is "S3Path"
y = cloudpathlib.CloudPath(x)
reveal_type(y)  # Pylance: Type of "y" is "S3Path"
```

These are the mypy bugs that arise:
```console
$ mypy cloudpathlib
cloudpathlib/cloudpath.py:127: error: Too many arguments for "__new__" of "object"  [call-arg]
cloudpathlib/cloudpath.py:129: error: Value of type variable "Self" of "__init__" of "CloudPath" cannot be "object"  [type-var]
cloudpathlib/cloudpath.py:130: error: Incompatible return value type (got "CloudPath", expected "CloudPathT")  [return-value]
cloudpathlib/cloudpath.py:141: error: Argument 1 to "__new__" of "ABCMeta" has incompatible type "CloudPathMeta"; expected "Type[<nothing>]"  [arg-type]
cloudpathlib/cloudpath.py:141: error: Argument 2 to "__new__" of "ABCMeta" has incompatible type "Union[str, CloudPathT]"; expected "str"  [arg-type]
```

I'm really not sure how to begin to solve these... it seems like they were there to begin with but weren't being raised (due to the lack of typing present on the signature, I assume)... I could definitely use some help in solving these. Would we consider `type: ignore`?

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -98,10 +98,11 @@ class CloudImplementation:
 implementation_registry: Dict[str, CloudImplementation] = defaultdict(CloudImplementation)
 
 
-def register_path_class(key: str) -> Callable:
-    T = TypeVar("T", bound=Type[CloudPath])
+CloudPathT = TypeVar("CloudPathT", bound="CloudPath")
 
-    def decorator(cls: Type[T]) -> Type[T]:
+
+def register_path_class(key: str) -> Callable[[Type[CloudPathT]], Type[CloudPathT]]:
+    def decorator(cls: Type[CloudPathT]) -> Type[CloudPathT]:
         if not issubclass(cls, CloudPath):
             raise TypeError("Only subclasses of CloudPath can be registered.")
         implementation_registry[key]._path_class = cls
@@ -112,7 +113,7 @@ def register_path_class(key: str) -> Callable:
 
 
 class CloudPathMeta(abc.ABCMeta):
-    def __call__(cls, cloud_path, *args, **kwargs):
+    def __call__(cls, cloud_path: Union[str, CloudPathT], *args, **kwargs) -> CloudPathT:
         # cls is a class that is the instance of this metaclass, e.g., CloudPath
 
         # Dispatch to subclass if base CloudPath


ringohoffman:
@jayqi @pjbull

pjbull:
@ringohoffman Thanks for looking in to this. Are these new problems because of #327? Do things work as expected on an earlier release? Looking at that PR may help with the mypy issues you are seeing.

ringohoffman:
No, the errors are there for `pyright==1.1.307` even on the commit immediately before #327 (a0dd9275f22fd2904ce9f9d36038c821fc22d2ca).

ringohoffman:
@pjbull let me know how this works for you, it seems to give us the better behavior than we saw before:

```python
import cloudpathlib


# pyright 1.1.306 (master)
x = cloudpathlib.S3Path("s3://bucket/key")
# reveal_type(x)  # pyright:Type of "x" is "S3Path"
y = cloudpathlib.CloudPath(x)
# reveal_type(y)  # pyright: Type of "y" is "CloudPath"
z = cloudpathlib.CloudPath("s3://bucket/key")
# reveal_type(z)  # pyright: Type of "z" is "CloudPath"

# pyright 1.1.307 (master)
x = cloudpathlib.S3Path("s3://bucket/key")
# reveal_type(x)  # pyright: Type of "x" is "CloudPath | Unknown | Self@__new__"
# reveal_type(x)  # mypy: Revealed type is "cloudpathlib.s3.s3path.S3Path"
y = cloudpathlib.CloudPath(x)
# reveal_type(y)  # pyright: Type of "y" is "CloudPath | Unknown | Self@__new__"
# reveal_type(y)  # mypy: Revealed type is "cloudpathlib.cloudpath.CloudPath"
z = cloudpathlib.CloudPath("s3://bucket/key")
# reveal_type(z)  # pyright: Type of "z" is "CloudPath | Unknown | Self@__new__"
# reveal_type(z)  # mypy: Revealed type is "cloudpathlib.cloudpath.CloudPath"

# pyright 1.1.307 (this PR)
x = cloudpathlib.S3Path("s3://bucket/key")
# reveal_type(x)  # pyright: Type of "x" is "S3Path"
# reveal_type(x)  # mypy: Revealed type is "cloudpathlib.s3.s3path.S3Path"
y = cloudpathlib.CloudPath(x)
# reveal_type(y)  # pyright: Type of "y" is "S3Path"
# reveal_type(y)  # mypy: Revealed type is "cloudpathlib.cloudpath.CloudPath"
z = cloudpathlib.CloudPath("s3://bucket/key")
# reveal_type(z)  # pyright: Type of "z" is "CloudPath"
# reveal_type(z)  # mypy: Revealed type is "cloudpathlib.cloudpath.CloudPath"

```

It's probably worth pointing out that `mypy` throws errors in all cases for `y` and `z`:

```
Cannot instantiate abstract class "CloudPath" with abstract attributes "drive", "is_dir", "is_file", "mkdir" and "touch"  [abstract]mypy(error)
```

I do feel like this pattern is pretty abnormal and I wonder if we couldn't employ a more standard design that doesn't involve this custom metaclass. Alternate constructor `classmethod`s might be a slightly more standard and better supported approach: https://stackoverflow.com/a/141777/7059681. We could also write a decorator to copy the docstrings of `pathlib` methods as well.

cloudpathlib/cloudpath.py
```diff
@@ -135,12 +135,10 @@ def __call__(cls, cloud_path, *args, **kwargs):
             raise InvalidPrefixError(
                 f"Path {cloud_path} does not begin with a known prefix {valid}."
             )
-
-        # Otherwise instantiate as normal
-        new_obj = cls.__new__(cls, cloud_path, *args, **kwargs)
-        if isinstance(new_obj, cls):
-            cls.__init__(new_obj, cloud_path, *args, **kwargs)
-        return new_obj
+        assert issubclass(cls, CloudPath)
```
pjbull, line 138: Let's make this an explicit check and raise rather than an assert.

pjbull (changes requested):
Thanks @ringohoffman! This change looks good except removing one `assert`. When that is done, I'll bring it over to a local branch to run the cloud tests.

#328 is tracking the `mypy` issue as you note, which seems like an actual bug with `mypy`: https://github.com/python/mypy/issues/14122 🤞 for an upstream fix there.

> I do feel like this pattern is pretty abnormal and I wonder if we couldn't employ a more standard design that doesn't involve this custom metaclass. Alternate constructor classmethods might be a slightly more standard and better supported approach: https://stackoverflow.com/a/141777/7059681. We could also write a decorator to copy the docstrings of pathlib methods as well.

May be worth opening an issue to discuss this, but my general opinion is that we shouldn't sacrifice what feels like a ergonomic API design because of type hinting. It's nice to have `CloudPath(p)` dispatch to the right subclass no matter the value of `p`, and I wouldn't be in favor of changing it just because the type hinting can go wonky. There's no implementation in which a static type checker could ever know the dispatched class at runtime in this example. If people are only dealing with one provider, I would recommend they use the provider-specific class (e.g., `S3Path`) if they want type checking guarantees. Otherwise, best case scenario is they get the generic `CloudPath` hints.

This use case is the reason that metaclasses exist as a language feature, so I don't think it is really abnormal—just a tension between features for a dynamic language and static type hinting. I'm not sure there's another implementation that let's us have our cake and eat it too, but open to discussing options if there are.

ringohoffman:
@pjbull I went ahead and applied your suggestion. I also found this example of metaclass hinting that gives the same behavior as before without any needing any `type: ignore` comments: https://mypy.readthedocs.io/en/stable/metaclasses.html#metaclass-usage-example

I think the latest way is the best way to hint this method.

ringohoffman:
@pjbull any thoughts about this latest change?

pjbull (approved):

PR #332, opened 2023-05-11
Allow empty content type for s3 objects

### LINKED ISSUE #331: S3 object does not always have a content type
Cloudpathlib  expects all s3 objects to have a content type, however this is not always the case as `content-type` is not a required field and is part of the objects optional metadata. Please review https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#object-metadata

Please review the below code:
https://github.com/drivendataorg/cloudpathlib/blob/28b2eb5dbba3ba5477284e1b432ac723f1a87f1b/cloudpathlib/s3/s3client.py#LL145C9-L145C49

This will result in a `KeyError` on that specific line.

From what I have seen, the content type is not even used by cloudpathlib, so my vote would be to ditch it.

georgeboot:
Fixes #331

### DIFF
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -142,7 +142,7 @@ class S3Client(Client):
             "last_modified": data["LastModified"],
             "size": data["ContentLength"],
             "etag": data["ETag"],
-            "content_type": data["ContentType"],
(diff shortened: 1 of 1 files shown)


jayqi (approved):
Thanks!

PR #334, opened 2023-05-17
Stream to and from Azure

georgeboot:
Currently, Azure is implemented without streams. This means that when uploading a file, the full file is first read into memory, before it is actually uploaded. The same goes for downloads.

This PR changes that by using streams.

### DIFF
--- a/cloudpathlib/azure/azblobclient.py
+++ b/cloudpathlib/azure/azblobclient.py
@@ -266,7 +266,8 @@ class AzureBlobClient(Client):
 
         content_settings = ContentSettings(**extra_args)
 
-        blob.upload_blob(Path(local_path).read_bytes(), overwrite=True, content_settings=content_settings)  # type: ignore
+        with open(Path(local_path), "rb") as data:
+            blob.upload_blob(data, overwrite=True, content_settings=content_settings)  # type: ignore
 
         return cloud_path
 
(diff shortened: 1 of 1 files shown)


pjbull:
@georgeboot Thanks, I'm interested in including something like this. A few questions/needs:

 - (1) Can we make similar changes for other backends? We generally like to implement things at parity.
 - (2) What happens on network failures? Is this better/worse than the current behavior? E.g., can we end up with partial files in the cache that then confuse future checks because they exist and the modified time is right, but the content is wrong?
 - (3) Passing tests

georgeboot:
Sure, will have a look!

S3 uses streams internally already. GCP I'll have to check.

When a network error happens, there indeed is a partial file on disk. If you want, I can clean that up? What should happen if we are overwriting an existing file and we fail half way through?

Will fix tests as well. Wanted to see if this had a chance to get merged first.

pjbull:
> What should happen if we are overwriting an existing file and we fail half way through?

Maybe this should append `.partial` to the end of the local file until download completes successfully and then rename? This is maybe annoying in that you need potentially 2x the file size for the cache, but it seems like a good balance of simple to implement and reliable.

pjbull:
Hi @georgeboot, this change still seems worth having. Would you have a chance to wrap what this PR needs? I think that should be:

 - [ ] Rebase on to lastest so unrelated tests don't fail
 - [ ] Management of partially downloaded files as noted above
 - [ ] Confirm GCS also streams
 - [ ] Tests as needed (particularly, the partially downloaded files logic)

pjbull:
Merging in to local branch to finish implementation.

PR #338, opened 2023-06-12
Make CloudPath.is_valid_cloudpath a TypeGuard (#337)

pjbull:
This allows us to infer that CloudPaths are the same type of object as the class is_valid_cloudpath is called on

------------------

Local version of #337 for live tests


ringohoffman:
Could we release a patch for this + #336? I'd really like to consume them!

Issue #179, opened 2021-11-11
Pylance shows typing errors


karolzlot:
For example I need to use `str()` for `.stem` field, otherwise it shows:

![image](https://user-images.githubusercontent.com/11590960/141295834-d8171fdd-5b07-46bd-acc0-c4db1ada5251.png)

pjbull:
Pylance uses `pyright` to do type checking. We use (and pass) with `mypy`. Here's the error output with `pyright` for reference if anyone wants to tackly making pyright pass as well:

```
❯ pyright cloudpathlib
npx: installed 1 in 1.553s
No configuration file found.
pyproject.toml file found at /Users/bull/code/cloudpathlib.
Loading pyproject.toml file at /Users/bull/code/cloudpathlib/pyproject.toml
Pyproject file "/Users/bull/code/cloudpathlib/pyproject.toml" is missing "[tool.pyright] section.
stubPath /Users/bull/code/cloudpathlib/typings is not a valid directory.
Assuming Python platform Darwin
Searching for source files
Found 25 source files
/Users/bull/code/cloudpathlib/cloudpathlib/client.py
  /Users/bull/code/cloudpathlib/cloudpathlib/client.py:58:16 - error: Object of type "None" cannot be called (reportOptionalCall)
/Users/bull/code/cloudpathlib/cloudpathlib/cloudpath.prefixes.py
  /Users/bull/code/cloudpathlib/cloudpathlib/cloudpath.prefixes.py:66:64 - error: Cannot access member "__name__" for type "None"
    Member "__name__" is unknown (reportGeneralTypeIssues)
  /Users/bull/code/cloudpathlib/cloudpathlib/cloudpath.prefixes.py:89:38 - error: Cannot assign member "_path_class" for type "CloudImplementation"
    Expression of type "Type[CloudPath]" cannot be assigned to member "_path_class" of class "CloudImplementation"
      Type cannot be assigned to type "None" (reportGeneralTypeIssues)
  /Users/bull/code/cloudpathlib/cloudpathlib/cloudpath.prefixes.py:122:31 - error: Argument of type "Self@CloudPathMeta" cannot be assigned to parameter "cls" of type "Type[_TT@__new__]" in function "__new__"
    Type "Self@CloudPathMeta" cannot be assigned to type "Type[_TT@__new__]" (reportGeneralTypeIssues)
[...]

PR #342, opened 2023-06-18
Add "CloudPath" to possible `cloud_path` types in `CloudPath` class

karolzlot:
This solves my issue in https://github.com/drivendataorg/cloudpathlib/issues/179#issuecomment-1595921134

After reading https://peps.python.org/pep-0673/ it seems to me that this change won't cause any problems. What do you think?


karolzlot:
It seems to me that "Self" is misused here so we could also remove it.

pjbull:
Merging into a local branch to run live server tests

PR #348, opened 2023-07-09
pydantic compatibility warning

pjbull:
Actual migration to support 2.0.0 is going to be a bunch more work, so adding just the warning for the time being.

```python
❯ pip list | grep pydantic
pydantic                   1.10.11

❯python
Python 3.11.4 (main, Jul  5 2023, 08:41:25) [Clang 14.0.6 ] on darwin
>>> import cloudpathlib
>>> quit()

❯ pip install -U pydantic
...
Successfully installed pydantic-2.0.2

❯ python
Python 3.11.4 (main, Jul  5 2023, 08:41:25) [Clang 14.0.6 ] on darwin
>>> import cloudpathlib
/Users/bull/code/cloudpathlib/cloudpathlib/__init__.py:49: PydanticVersionWarning: This version of cloudpathlib is only compatible with pydantic<2.0.0. You can ignore this warning if none of your pydantic models are annotated with cloudpathlib types.
  warn(PydanticVersionWarning(PydanticVersionWarning.message))
```

Docs preview:
https://deploy-preview-348--gallant-agnesi-5f7bb3.netlify.app/integrations/
image

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,9 @@
 # cloudpathlib Changelog
 
+## v0.15.1 (2023-06-16)
+ - Warn if Pydantic >= 2.0.0 since we are not yet compatible.
+
+
 ## v0.15.0 (2023-06-16)
 
 - Changed return type for `CloudPathMeta.__call__` to fix problems with pyright/pylance ([PR #330](https://github.com/drivendataorg/cloudpathlib/pull/330))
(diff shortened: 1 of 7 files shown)


pjbull:
Actually, not sure that we need this PR if we merge #349. It may be nice to give folks a warning on this version, but since we can do side-by-side, people can simply upgrade to the next release if they need pydantic v2 support.

PR #349, opened 2023-07-10
Pydantic v2 update

pjbull:
~~Note: This is on top of #348 and should be merged after that. Commits from `65d7911` have the relevant changes.~~

**Updated to be an incremental change without needing different pydantic v1/v2 compatible versions.**

Update for compatibility with pydantic v2. Since the two relevant method names changed entirely, we can keep v1/v2 compatibility side by side.

Manually ran test suite on pydantic 1.10 and 2.0

Tried a few other implementations [from the docs](https://docs.pydantic.dev/2.0/usage/types/custom/) like using `Annotated` and also a separate mixin class, but ultimately this implementation, which is a minimal set of changes from our v1 implementation, is the one that played best with our dispatch/polymorphism (and didn't require any public API changes).

### DIFF
--- a/tests/mock_clients/mock_s3.py
+++ b/tests/mock_clients/mock_s3.py
@@ -90,6 +90,9 @@ class MockBoto3Object:
 
     def download_file(self, to_path, Config=None, ExtraArgs=None):
         to_path = Path(to_path)
+
+        to_path.parent.mkdir(parents=True, exist_ok=True)
+
         to_path.write_bytes(self.path.read_bytes())
         # track config to make sure it's used in tests
         self.resource.download_config = Config
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,9 @@
 # cloudpathlib Changelog
 
+## v0.15.1 (2023-06-16)
+ - Warn if Pydantic >= 2.0.0 since we are not yet compatible.
+
+
 ## v0.15.0 (2023-06-16)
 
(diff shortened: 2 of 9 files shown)


tests/mock_clients/mock_s3.py
```diff
@@ -90,6 +90,9 @@ def copy_from(self, CopySource=None, Metadata=None, MetadataDirective=None):
 
     def download_file(self, to_path, Config=None, ExtraArgs=None):
         to_path = Path(to_path)
+
+        to_path.parent.mkdir(parents=True, exist_ok=True)
```
pjbull, line 94: Added to address MacOS test failures:
https://github.com/drivendataorg/cloudpathlib/actions/runs/5366826109/jobs/9736473309

I couldn't get a root cause why that was the only scenario of cloud provider/os to fail and why it started happening without a code change on our part. We actually already do this creation in the Mock for GCS, and if it were a problem with the test code not making all the directories in the ways end users would need to, I would expect the live tests to fail.

jayqi (approved):
Looks good. I wonder if we could get the Annotated version of this to work, because I recall that adding these methods makes `AnyPath` not work as a protocol class for type checking with `CloudPath` and `Path`.

pjbull:
> Looks good. I wonder if we could get the Annotated version of this to work, because I recall that adding these methods makes AnyPath not work as a protocol class for type checking with CloudPath and Path.

Yeah, it does seem like it should be possible, but IIRC it wasn't playing nice with the current `ABC` implementation of `AnyPath`.

PR #361, opened 2023-08-20
Implement env var for local cache dir

### LINKED ISSUE #352: Set local_cache_dir from environment variable?
If I want a persistent cache, I can set `CLOUPATHLIB_FILE_CACHE_MODE`. But I then have to pass a `local_cache_dir` in code - why not allow this to also be set from the environment?

pjbull:
Closes #352


tests/test_caching.py
```diff
@@ -278,6 +279,28 @@ def test_environment_variable_instantiation(rig: CloudProviderTestRig, tmpdir):
         os.environ["CLOUPATHLIB_FILE_CACHE_MODE"] = original_env_setting
 
 
+def test_environment_variable_local_cache_dir(rig: CloudProviderTestRig, tmpdir):
```
jayqi, line 282: `tmpdir` is deprecated and we should use `tmp_path` instead: https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

jayqi (changes requested):

tests/test_caching.py
pjbull: Added issue #370 to make this change everywhere

PR #365, opened 2023-09-28
PR: fix localpath fallback

hoshimura:
I would really like to see cloudpathlib as a transparent replacement to pathlib and with this PR I have tried to mitigate the problems I encountered when the path type is not known and there might be a mix between local and remote paths.
Looking forward to comments on where I rounded to many corners.
Change-Id: I19d3e43096dbc46671ad55d28dadc7741d09822d

### DIFF
--- a/cloudpathlib/local/localclient.py
+++ b/cloudpathlib/local/localclient.py
@@ -7,11 +7,11 @@ import shutil
 from tempfile import TemporaryDirectory
 from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 
-from ..client import Client
+from ..client import Client, register_client_class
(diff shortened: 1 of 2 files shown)


pjbull:
@hoshimura Could you file an issue with some more details about what you're trying to do and what is and isn't working?

hoshimura:
Ok, I guess I am more of a doer than a reader. It was obviously AnyPath I was looking for. Just confused why it wasn't under the CloudPath umbrella namewise. My problem was that I tried CloudPath("localpath") and got InvalidPrefixError. Not finding the obvious, but instead the LocalPath and client.... Ok, apart from this PR handles file:// uris everything I wish for with this PR is in AnyPath

PR #373, opened 2023-10-30
Fix an error when incomplete S3Client is destroyed

bryanwweber:
If an error occurs during initialization of the S3Client instance, the initializer for the Client superclass is never called. This could happen, for example, when the S3Client has a bad profile name passed which botocore doesn't understand. The specific error is that the file_cache_mode attribute is not set because the superclass initializer is not called in the S3Client initializer if an error occurs during S3 session setup. This commit moves the superclass initializer to the top of the S3Client initializer to ensure that all superclass attributes are defined if/when any errors occur.

The same fix is applied for the Google client and the local client.

Closes #372

----------------

Contributor checklist:

 - [X] I have read and understood `CONTRIBUTING.md`
 - [X] Confirmed an issue exists for the PR, and the text `Closes #issue` appears in the PR summary (e.g., `Closes #123`).
 - [X] Confirmed PR is rebased onto the latest base
 - [x] Confirmed failure before change and success after change
 - [X] Any generic new functionality is replicated across cloud providers if necessary
 - [ ] Tested manually against live server backend for at least one provider
 - [ ] Added tests for any new functionality
 - [ ] Linting passes locally
 - [ ] Tests pass locally
 - [x] Updated `HISTORY.md` with the issue that is addressed and the PR you are submitting. If the top section is not `## UNRELEASED``, then you need to add a new section to the top of the document for your change.

PR #375, opened 2023-11-01
Fix an error when incomplete S3Client is destroyed (#373)

### LINKED ISSUE #372: `AttributeError` when __del__ is called on S3Client that errors during __init__
If an exception is raised during the creation of an S3Client instance, subsequently tearing down that instance via `Client.__del__` results n an `AttributeError`:

```python
❯ poetry run python        
Python 3.11.5 (main, Sep 22 2023, 12:39:47) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin
>>> import cloudpathlib
>>> cloudpathlib.S3Client(profile_name="some-bad-profile")
  File "<stdin>", line 1, in <module>
  File ".venv/lib/python3.11/site-packages/cloudpathlib/s3/s3client.py", line 86, in __init__
    self.sess = Session(
                ^^^^^^^^
  File "venv/lib/python3.11/site-packages/boto3/session.py", line 90, in __init__
    self._setup_loader()
  File ".venv/lib/python3.11/site-packages/boto3/session.py", line 131, in _setup_loader
    self._loader = self._session.get_component('data_loader')
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/session.py", line 802, in get_component
    return self._components.get_component(name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/session.py", line 1140, in get_component
    self._components[name] = factory()
                             ^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/session.py", line 199, in <lambda>
    lambda: create_loader(self.get_config_variable('data_path')),
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/botocore/session.py", line 323, in get_config_variable
    return self.get_component('config_store').get_config_variable(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/configprovider.py", line 459, in get_config_v
ariable
    return provider.provide()
           ^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/configprovider.py", line 665, in provide
    value = provider.provide()
            ^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/configprovider.py", line 755, in provide
    scoped_config = self._session.get_scoped_config()
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/session.py", line 422, in get_scoped_config
    raise ProfileNotFound(profile=profile_name)
botocore.exceptions.ProfileNotFound: The config profile (some-bad-profile) could not be found
>>> quit()
Exception ignored in: <function Client.__del__ at 0x1031bcc20>
  File ".venv/lib/python3.11/site-packages/cloudpathlib/client.py", line 88, in __del__
    if self.file_cache_mode in [
       ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'S3Client' object has no attribute 'file_cache_mode'
```

This is happening because the `super().__init__()` call happens at the bottom of the `S3Client.__init__()` method ([here](https://github.com/drivendataorg/cloudpathlib/blob/ffbdd1c7af1e90bcfdabb1432f5982be31263e13/cloudpathlib/s3/s3client.py#L130-L134)), so `Client.file_cache_mode` is not set before the error occurs. I checked and there doesn't seem to be a particular reason that `super().__init__()` needs to be called at the bottom (all the args are just pass through from the `S3Client.__init__()` signature). I'm happy to submit a PR to fix this by moving the `super().__init__()` up in `S3Client.__init__()` and other related classes too, if that seems like the right fix. Thanks!

pjbull:
Repo-local PR for live tests of #373

------------

* Fix an error when incomplete S3Client is destroyed

If an error occurs during initialization of the S3Client instance, the initializer for the Client superclass is never called. This could happen, for example, when the S3Client has a bad profile name passed which botocore doesn't understand. When the remainder of the class is garbage collected (for example, at interpreter shutdown), Python calls the __del__ method of the Client superclass. The specific error is that the file_cache_mode attribute is not set because the superclass initializer is not called in the S3Client initializer if an error occurs during S3 session setup.

This change uses getattr with a default value to avoid the AttributeError in the __del__ method.

Closes #372

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,8 +1,11 @@
 # cloudpathlib Changelog
 
+## Unreleased
+- Fix `S3Client` cleanup via `Client.__del__` when `S3Client` encounters an exception during initialization. (Issue [#372](https://github.com/drivendataorg/cloudpathlib/issues/372), PR [#373](https://github.com/drivendataorg/cloudpathlib/pull/373))
+
 ## v0.16.0 (2023-10-09)
  - Add "CloudPath" as return type on `__init__` for mypy issues. ([Issue #179](https://github.com/drivendataorg/cloudpathlib/issues/179), [PR #342](https://github.com/drivendataorg/cloudpathlib/pull/342))
- - Add `with_stem` to all path types when python version supports it (>=3.9). ([Issue #287](https://github.com/drivendataorg/cloudpathlib/issues/287), [PR #290](https://github.com/drivendataorg/cloudpathlib/pull/290), thanks to [@Gilthans](https://github.com/Gilthans)) 
+ - Add `with_stem` to all path types when python version supports it (>=3.9). ([Issue #287](https://github.com/drivendataorg/cloudpathlib/issues/287), [PR #290](https://github.com/drivendataorg/cloudpathlib/pull/290), thanks to [@Gilthans](https://github.com/Gilthans))
  - Add `newline` parameter to the `write_text` method to align to `pathlib` functionality as of Python 3.10. [PR #362](https://github.com/drivendataorg/cloudpathlib/pull/362), thanks to [@pricemg](https://github.com/pricemg).
  - Add support for Python 3.12 ([PR #364](https://github.com/drivendataorg/cloudpathlib/pull/364))
  - Add `CLOUDPATHLIB_LOCAL_CACHE_DIR` env var for setting local_cache_dir default for clients ([Issue #352](https://github.com/drivendataorg/cloudpathlib/issues/352), [PR #357](https://github.com/drivendataorg/cloudpathlib/pull/357))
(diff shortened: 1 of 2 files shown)


pjbull:
Everything passes here, so I will merge. I think the windows test on #373 may just have been flaky. Thanks @bryanwweber!

PR #404, opened 2024-02-18
Implement file scheme

### LINKED ISSUE #401: Support file: ?
It's sympathic to use cloudpathlib with url `file:`.

pjbull:
Simple implementation of `file:` scheme in `AnyPath`.

Uses implementation from [here](https://github.com/python/cpython/issues/91504#issuecomment-1100257658), but it looks like there is a [`from_uri` method coming to `pathlib` in 3.13](https://github.com/python/cpython/issues/107465).

Closes #401


jayqi (commented):
Do we correctly handle percent-encoded colons from a drive on Windows? e.g.,

`file://c:/path/to/file.txt` vs. `file://c%3A/path/to/file.txt`

We should at least have a test for it. 

See this PR https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1664

that also referenced the cpython issue that you took this implementation from. They discuss the issue in that PR and have a Windows test case.

I tried to search around, and I found this which suggests that the percent-encoded version should be considered valid: https://github.com/microsoft/language-server-protocol/pull/1786/files

pjbull:
Updated, thanks @jayqi!

jayqi (approved):

PR #409, opened 2024-02-26
Version floor for google-cloud-storage

### LINKED ISSUE #408: What google-cloud-storage versions are supported?
The recent v0.18.0 release no longer works with google-cloud-storage 1.43, due to PR #391:

```
INFO:__main__:API:   File "/usr/local/lib/python3.11/site-packages/cloudpathlib/__init__.py", line 8, in <module>
INFO:__main__:API:     from .gs.gspath import GSPath
INFO:__main__:API:   File "/usr/local/lib/python3.11/site-packages/cloudpathlib/gs/__init__.py", line 1, in <module>
INFO:__main__:API:     from .gsclient import GSClient
INFO:__main__:API:   File "/usr/local/lib/python3.11/site-packages/cloudpathlib/gs/gsclient.py", line 19, in <module>
INFO:__main__:API:     from google.cloud.storage import transfer_manager
INFO:__main__:API: ImportError: cannot import name 'transfer_manager' from 'google.cloud.storage' (/usr/local/lib/python3.11/site-packages/google/cloud/storage/__init__.py)
```

This old version of google-cloud-storage does not provide `transfer_manager`, which was presumably added in 2.x.

Previous cloudpathlib releases did work with this old google-cloud-storage, or at least they worked well enough to import successfully and for basic functionality to work.

Admittedly it's silly of us to still be using this old version and we should probably update our google-cloud-storage requirements, but I didn't notice any required version specified in _pyproject.toml_ (whereas there is a listed minimum for other backends) or elsewhere:

```toml
[project.optional-dependencies]
azure = ["azure-storage-blob>=12"]
gs = ["google-cloud-storage"]
s3 = ["boto3>=1.34.0"]
```

Does cloudpathlib have a recommended or required minimal version for google-cloud-storage?

jayqi:
Fixes an import error when cloudpathlib is installed alongside an incompatible version of google-cloud-storage.

I think the plan here should be:

- Merge this into branch `hotfix/v0.18.1-google-cloud-storage` which we tag and maintain as the source for a v0.18.1 release.
- Yank v0.18.0, so that resolvers will not install v0.18.0 alongside `google-cloud-storage<2.7.0`

Closes #408.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,7 +2,7 @@
 
 ## v0.18.1 (2024-02-26)
 
-- Fixed import error due to incompatible google-cloud-strage by adding a version floor. ([Issue #408](https://github.com/drivendataorg/cloudpathlib/issues/408)) 
+- Fixed import error due to incompatible google-cloud-strage by adding a version floor. ([Issue #408](https://github.com/drivendataorg/cloudpathlib/issues/408), (PR #409)[https://github.com/drivendataorg/cloudpathlib/pull/409]) 
 
 **Note: This is the last planned Python 3.7 compatible release version.**
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,8 @@
 
 - Fixed import error due to incompatible google-cloud-strage by adding a version floor. ([Issue #408](https://github.com/drivendataorg/cloudpathlib/issues/408), (PR #409)[https://github.com/drivendataorg/cloudpathlib/pull/409]) 
 
+Includes all changes from v0.18.0.
+
 **Note: This is the last planned Python 3.7 compatible release version.**
 
 ## 0.18.0 (2024-02-25) (Yanked)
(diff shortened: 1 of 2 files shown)


pjbull:
@jayqi I had mocked a slightly different fix to support older versions like this: https://github.com/drivendataorg/cloudpathlib/commit/7e3a4696f41d239dd74fe6c86a3616df93d2343c

Thoughts?

jayqi:
Supporting both cases is more complexity so I have a weak preference against it—I recognize that your fix is not hugely complicated. I do think we'd also want to log a warning or something though, if someone tries to use it and it ends up not happening. google-cloud-storage v2.6.0 is from November 2022. 

Either way, I think we'll want to yank v0.18.0,

pjbull:
Yeah, agreed the release/yank plan looks good, thanks for figuring that out.

The only reason I have a slight preference for supporting both is that it seems that at least a couple people hit this pretty quickly so there may be a fair number of systems on older `google-cloud-storage` versions (and the fix is pretty isolated).

jayqi:
I'm okay with that. Feel free to open a new branch/PR off of [`hotfix/v0.18.1-google-cloud-storage`](https://github.com/drivendataorg/cloudpathlib/tree/hotfix/v0.18.1-google-cloud-storage) and borrow whatever changes from this PR seem worthwhile to keep.

jayqi:
Closed in favor of #410.

PR #410, opened 2024-02-26
Support versions of GCS without transfer_manager

pjbull:
Add support for `google-cloud-storage<2.7.0` per discussion in #408 and #409.

- Manually confirmed the test suite passes with `google-cloud-storage==2.6.0` (and fails without the fix)
- Warns if user passes `transfer_manager` kwargs that can't be used
- Adds changelog notes
- Bumps version

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,12 +1,21 @@
 # cloudpathlib Changelog
 
-## 0.18.0 (2024-02-25)
+## v0.18.1 (2024-02-26)
+
+- Fixed import error due to incompatible google-cloud-strage by adding a version floor. ([Issue #408](https://github.com/drivendataorg/cloudpathlib/issues/408), (PR #409)[https://github.com/drivendataorg/cloudpathlib/pull/409]) 
+
+Includes all changes from v0.18.0.
+
+**Note: This is the last planned Python 3.7 compatible release version.**
+
+## 0.18.0 (2024-02-25) (Yanked)
+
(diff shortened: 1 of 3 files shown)


HISTORY.md
```diff
@@ -1,12 +1,21 @@
 # cloudpathlib Changelog
 
-## 0.18.0 (2024-02-25)
+## v0.18.1 (2024-02-26)
+
+- Fixed import error due to incompatible google-cloud-strage by adding a version floor. ([Issue #408](https://github.com/drivendataorg/cloudpathlib/issues/408), (PR #410)[https://github.com/drivendataorg/cloudpathlib/pull/410])
```
jmarshall, line 5: Thanks, this all looks good for our purposes. Minor nit: “by adding a version floor” is probably no longer an accurate description of this version of the fix.

Also typo:
```suggestion
- Fixed import error due to incompatible google-cloud-storage by adding a version floor. ([Issue #408](https://github.com/drivendataorg/cloudpathlib/issues/408), (PR #410)[https://github.com/drivendataorg/cloudpathlib/pull/410]) 
```

jayqi:
I guess now I'm wondering if yanking v0.18.0 is necessary. It's a little hard to tell how severe of an action yanking is, and most of the examples given are worse errors. 

In the approach where we have a version floor, it's definitely necessary or v0.18.0 would be the newest version of cloudpathlib for `google-cloud-storage<2.7.0`. In this case, v0.18.1 would just be newer than v0.18.0 in all cases. 

I guess since this can cause an error on import just as a side-effect of having both packages installed, it probably doesn't hurt to still yank.

jayqi (approved):

PR #412, opened 2024-03-01
Use uv for CI

jayqi:
Swaps in [drivendataorg/setup-python-uv-action](https://github.com/drivendataorg/setup-python-uv-action) and installs our dependencies with uv (without caching). Seems to shave ~4 minutes off of our CI.


pjbull (approved):
Heck yah!

Issue #356, opened 2023-07-25
exists(gs://bucket/) fails with StopIteration on an empty bucket 


indigoviolet:
See previously: #291 , where it seems to have been fixed for S3

```
from cloudpathlib import CloudPath
CloudPath("gs://empty-bucket/").exists() # replace bucket name
```

pjbull:
Thanks for the issue, looks like the implementation doesn't work right for empty buckets since it uses iteration:
https://github.com/drivendataorg/cloudpathlib/blob/45ae4ee163c06ef2a2e5939ede29a84630a4e08e/cloudpathlib/gs/gsclient.py#L147-L156

We should just use the `.exists` method on the bucket instead:
https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.bucket.Bucket#google_cloud_storage_bucket_Bucket_exists

Happy to accept a PR for this if you have the chance.

PR #420, opened 2024-04-04
fix: use native `exists()` method in `GSClient`

bachya:
SSIA

Closes https://github.com/drivendataorg/cloudpathlib/issues/356

----------------

Contributor checklist:

 - [ ] Any generic new functionality is replicated across cloud providers if necessary
 - [ ] Added tests for any new functionality
 - [x] Linting passes locally
 - [x] Tests pass locally

### DIFF
--- a/cloudpathlib/gs/gsclient.py
+++ b/cloudpathlib/gs/gsclient.py
@@ -14,7 +14,6 @@ try:
     if TYPE_CHECKING:
         from google.auth.credentials import Credentials
 
-    from google.api_core.exceptions import NotFound
     from google.auth.exceptions import DefaultCredentialsError
     from google.cloud.storage import Client as StorageClient
 
--- a/cloudpathlib/gs/gsclient.py
+++ b/cloudpathlib/gs/gsclient.py
@@ -169,7 +169,11 @@ class GSClient(Client):
                 return None
 
     def _exists(self, cloud_path: GSPath) -> bool:
-        return self.client.bucket(cloud_path.bucket).exists()
+        # short-circuit the root-level bucket
+        if not cloud_path.blob:
+            return self.client.bucket(cloud_path.bucket).exists()
+
+        return self._is_file_or_dir(cloud_path) in ["file", "dir"]
 
     def _list_dir(self, cloud_path: GSPath, recursive=False) -> Iterable[Tuple[GSPath, bool]]:
         # shortcut if listing all available buckets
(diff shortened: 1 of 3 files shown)


cloudpathlib/gs/gsclient.py
```diff
@@ -169,15 +169,7 @@ def _is_file_or_dir(self, cloud_path: GSPath) -> Optional[str]:
...
-        # short-circuit the root-level bucket
-        if not cloud_path.blob:
-            try:
-                next(self.client.bucket(cloud_path.bucket).list_blobs())
-                return True
-            except NotFound:
-                return False
-
-        return self._is_file_or_dir(cloud_path) in ["file", "dir"]
+        return self.client.bucket(cloud_path.bucket).exists()
```
pjbull, line 172: The `_exists()` is for all potential paths, not just buckets so I think the update is just for the short-circuit path of checking if the bucket exists.

Should be something like:

```python
if not cloud_path.blob:
    return self.client.bucket(cloud_path.bucket).exists()
    
return self._is_file_or_dir(cloud_path) in ["file", "dir"]
```

pjbull (changes requested):
Thanks @bachya!

One change here. Can you also confirm that you tested the case in #356 with a totally empty bucket and this method works?

bachya:
@pjbull Confirmed, I have this working on an empty GCP bucket.

pjbull:
Thanks @bachya, happy to merge this once the code quality and mocked tests are passing.

pjbull:
Tests need an update in the mocked version (which mock the cloud SDKs):
 - Update [`MockBucket` to have exists method.](https://github.com/drivendataorg/cloudpathlib/blob/c9da3f3ed7484f621f3ca33eef94ffcf0f0ca26c/tests/mock_clients/mock_gs.py#L120-L160)
 - Should be [implemented like the Azure backend to just be true for the test](https://github.com/drivendataorg/cloudpathlib/blob/c9da3f3ed7484f621f3ca33eef94ffcf0f0ca26c/tests/mock_clients/mock_azureblob.py#L142-L146).

More info on running the test suite here: https://cloudpathlib.drivendata.org/stable/contributing/#tests

bachya:
Thanks, @pjbull. I couldn't see the results until you approved the CI run; got them now, and I'll update the tests.

pjbull:
@bachya just fyi, you can run `make lint` and `make test` locally so you don't have to wait for the CI runs.

pjbull (approved):

pjbull:
Looks good! I'm going to merge into a repo-local branch so we can run against the live server backends, which is not possible on PRs from forks.

PR #424, opened 2024-04-08
Fixes typo in Caching envvar

### LINKED ISSUE #423: Typo in Cache Mode envvar
The cache mode envvar is `CLOUPATHLIB_FILE_CACHE_MODE` while it should be `CLOUDPATHLIB_FILE_CACHE_MODE`.

This is consistent across the project, however is definitely unexpected from a user perspective.

mynameisfiber:
Envvar to set caching mode was previously CLOUPATHLIB_FILE_CACHE_MODE (missing D in CLOUDPATHLIB). This commit fixes this typo and introduces some code to help migrate projects to the new envvar. The current logic for migration warnings is:

- If the old envvar is set and not the new one, use the old one and issue a deprecation warning
- If the old and new envvars are set, use the new version and issue a warning if the two disagree

Closes #423

----------------

Contributor checklist:

 - [ ] Tested manually against live server backend for at least one provider
 - [x] Linting passes locally
 - [x] Tests pass locally


cloudpathlib/enums.py
```diff
@@ -29,14 +30,33 @@ class FileCacheMode(str, Enum):
...
-        env_string = os.environ.get("CLOUPATHLIB_FILE_CACHE_MODE", "").lower()
 
+        env_string = os.environ.get("CLOUDPATHLIB_FILE_CACHE_MODE", "").lower()
+        env_string_typo = os.environ.get("CLOUPATHLIB_FILE_CACHE_MODE", "").lower()
+
+        if env_string_typo:
+            warnings.warn(
+                "envvar CLOUPATHLIB_FILE_CACHE_MODE has been renamed to "
+                "CLOUDPATHLIB_FILE_CACHE_MODE. Reading from the old value "
+                "will become deprecated in the future.",
```
pjbull, line 47: Let's say `deprecated in version 0.20.0` which gives a full minor release for people to move off of it.

pjbull (changes requested):
Great, thanks! I think this should be good when the tests are passing.

Could you also file an issue to remove the deprecated code in a future release?

pjbull (approved):

pjbull:
Moving to repo-local branch where the live backend tests can run. Thanks @mynameisfiber!

PR #433, opened 2024-05-03
Add exists check in download_to

### LINKED ISSUE #430: [GS] `download_to` with non-existent files don't raise any exception
Version: 0.18.1 

Code Example:
```py

# download a non-existent file, in an existing bucket
path = CloudPath("gs://some_existent_bucket/non_existent.txt")
path.download_to(".") # should raise an exception, as the file doesn't exist

# NOTE: trying to download from a non-existing bucket will properly fail
path = CloudPath("gs://non_existent_bucket/foo.txt")
path.download_to(".") # will raise an exception
```

pjbull:
Add `.exists` check before `download_to` and raise new `CloudPathNotExistsError` if the path does not exist.

Closes #430


jayqi:
This adds an extra network call, right? (And we're already doing multiple network calls even before this change.) I don't love this, but I'm willing to be convinced this is okay for now and we track an issue to do this more intelligently.

pjbull:
> This adds an extra network call, right? (And we're already doing multiple network calls even before this change.) I don't love this, but I'm willing to be convinced this is okay for now and we track an issue to do this more intelligently.

Yeah, it does. Where I landed was that I'd rather implement that through caching that kind of metadata intelligently at the client level so that the implementation here stays the same and network overhead is reduced.

Also, it is one call in each `download_to`, so that if you are doing a directory it doesn't happen for every file.

jayqi (approved):

PR #436, opened 2024-05-25
Fixes for LocalClient bugs: glob in local clients to match cloud clients, reset default storage work on default client

### LINKED ISSUE #414: reset_default_storage_dir gets ignored through get_default_client
```python
from cloudpathlib.local import LocalS3Path, LocalS3Client

path = LocalS3Path("s3://some/path/")
path.touch()
LocalS3Client.reset_default_storage_dir()
new_path = LocalS3Path("s3://some/path/")
assert not new_path.is_file() # fails
```

The reason is that `reset_default_storage_dir()` acts on the class attribute only and not on the instance attribute of the `LocalClient` in question. Because of `Client.get_default_client()` returning the same client instance for any path (saved as a class attribute), no new client gets instantiated to make use of the new directory. The tests don’t catch that because they create the clients explicitly.

### LINKED ISSUE #415: LocalS3Path doesn’t support glob on non-existing directories
The implementations of S3Path and LocalS3Path diverge in the following way which disqualifies using `LocalS3Path` as a testing tool.

```python
from cloudpathlib.s3 import S3Path

path = S3Path("s3://this/path/does/not/exist")
path.mkdir(parents=True) # does nothing (because S3 doesn't have directories)
print(list(path.glob("*"))) # returns []
```

```python
from cloudpathlib.local import LocalS3Path

path = LocalS3Path("s3://this/path/does/not/exist")
path.mkdir(parents=True) # again, does nothing
print(list(path.glob("*"))) # raises exception
```

To fix this in my own codebase, I added a check to `LocalS3Path.glob`  to return an empty generator when the path is not a directory. However, I’m unsure if this fixes the issue in all cases and there might be other issues caused by directories not existing in S3 but locally.

pjbull:
- `glob` for local clients should return empty list (to match cloud version) rather than raise exception.
 - Reset default client in `reset_default_storage_dir` so the new storage dir gets picked up because the default client will have to be recreated

Closes #415
Closes #414

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,17 +3,21 @@
 ## UNRELEASED
 
 - Fixes typo in `FileCacheMode` where values were being filled by envvar `CLOUPATHLIB_FILE_CACHE_MODE` instead of `CLOUDPATHLIB_FILE_CACHE_MODE`. (PR [#424](https://github.com/drivendataorg/cloudpathlib/pull/424)
-- Fix `CloudPath` cleanup via `CloudPath.__del__` when `Client` encounters an exception during initialization and does not create a `file_cache_mode` attribute. (Issue [#372](https://github.com/drivendataorg/cloudpathlib/issues/372), thanks to [@bryanwweber](https://github.com/bryanwweber)) 
+- Fix `CloudPath` cleanup via `CloudPath.__del__` when `Client` encounters an exception during initialization and does not create a `file_cache_mode` attribute. (Issue [#372](https://github.com/drivendataorg/cloudpathlib/issues/372), thanks to [@bryanwweber](https://github.com/bryanwweber))
 - Drop support for Python 3.7; pin minimal `boto3` version to Python 3.8+ versions. (PR [#407](https://github.com/drivendataorg/cloudpathlib/pull/407))
 - fix: use native `exists()` method in `GSClient`. (PR [#420](https://github.com/drivendataorg/cloudpathlib/pull/420))
 - Enhancement: lazy instantiation of default client (PR [#432](https://github.com/drivendataorg/cloudpathlib/issues/432), Issue [#428](https://github.com/drivendataorg/cloudpathlib/issues/428))
 - Adds existence check before downloading in `download_to` (Issue [#430](https://github.com/drivendataorg/cloudpathlib/issues/430), PR [#432](https://github.com/drivendataorg/cloudpathlib/pull/432))
 - Add env vars `CLOUDPATHLIB_FORCE_OVERWRITE_FROM_CLOUD` and `CLOUDPATHLIB_FORCE_OVERWRITE_TO_CLOUD`. (Issue [#393](https://github.com/drivendataorg/cloudpathlib/issues/393), PR [#437](https://github.com/drivendataorg/cloudpathlib/pull/437))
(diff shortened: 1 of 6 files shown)


HISTORY.md
```diff
@@ -8,6 +8,7 @@
 - fix: use native `exists()` method in `GSClient`. (PR [#420](https://github.com/drivendataorg/cloudpathlib/pull/420))
 - Enhancement: lazy instantiation of default client (PR [#432](https://github.com/drivendataorg/cloudpathlib/issues/432), Issue [#428](https://github.com/drivendataorg/cloudpathlib/issues/428))
 - Adds existence check before downloading in `download_to` (Issue [#430](https://github.com/drivendataorg/cloudpathlib/issues/430), PR [#432](https://github.com/drivendataorg/cloudpathlib/pull/432))
+- Have `glob` for local clients match behavior of cloud versions for parity in testing; make `reset_default_storage_dir` work with default client. (Issue [#414](https://github.com/drivendataorg/cloudpathlib/issues/414), Issue [#415](https://github.com/drivendataorg/cloudpathlib/issues/415), [PR #436](https://github.com/drivendataorg/cloudpathlib/pull/436))
```
jayqi, line 11: These seem like two different unrelated bugs, and we should split them up into two separate bullets? Also, we should clean up the notes in this section to be a little more consistent with [Keep a Changelog](https://keepachangelog.com/en/1.1.0/#how) verb conventions.

cloudpathlib/local/localclient.py
```diff
@@ -49,6 +49,10 @@ def get_default_storage_dir(cls) -> Path:
     @classmethod
     def reset_default_storage_dir(cls) -> Path:
         cls._default_storage_temp_dir = None
+
+        # Also reset default client so it gets recreated with new temp dir
+        cls._default_client = None
```
jayqi, line 54: I'm not convinced this is the right fix. 

It took me a while to remember what `reset_default_storage_dir` is supposed to do and why, and I had to [look at our docs](https://cloudpathlib.drivendata.org/stable/testing_mocked_cloudpathlib/#2-get-a-pathlibpath-object-that-points-to-the-local-storage-directory).

It feels weird to me that resetting the storage directory has a side effect of changing the default client. If someone had set a default client with non-default values (e.g., say they set a non-default `file_cache_mode`), then resetting the default client here would lead to unexpected changes later.

It kind of seems to me like the more sensible design is that the storage directory shouldn't be a class instance at all, and that it should be a client instance attribute (i.e., setting storage directory to a default value should be a method and not a property or attribute). And then for cleanup, rather than a `reset_default_storage_dir` method, there should be a `reset_default_client` method.

jayqi (changes requested):

pjbull:
Update this PR to do the following:
 - [ ] Get rid of `_default_storage_temp_dir`. Only use the per-instance `self._local_storage_dir`.
 - [ ] Deprecation warning for `reset_default_storage_dir`
 - [ ] New `reset_default_client` that can be used for test cleanup and clears the `._default_client` instance on the class.
 - [ ] Update documentation on recommendations for test cleanup.

pjbull:
@jayqi I think this is probably the best simple fix at the moment. Fixes both the reported issue and [a similar issue in this SO post](https://stackoverflow.com/questions/78839557/local-implementation-of-s3-bucket-in-cloudpathlib-not-resetting-default-storage?noredirect=1#comment139025105_78839557).

jayqi (approved):

PR #437, opened 2024-05-26
Env vars for force overwrite functions

### LINKED ISSUE #393: Set force_overwrite_from_cloud from environment variable?
Similar to #352, is it possible to set `force_overwrite_from_cloud` from an environment variable?

pjbull:
Add env vars that let you override things for forcing overwrites.

Closes #393

### DIFF
--- a/cloudpathlib/cloudpath.py
+++ b/cloudpathlib/cloudpath.py
@@ -315,7 +315,7 @@ class CloudPath(metaclass=CloudPathMeta):
 
     def __fspath__(self) -> str:
         if self.is_file():
-            self._refresh_cache(force_overwrite_from_cloud=False)
+            self._refresh_cache()
         return str(self._local)
 
     def __lt__(self, other: Any) -> bool:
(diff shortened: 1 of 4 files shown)


tests/test_caching.py
```diff
@@ -344,6 +344,41 @@ def test_environment_variable_local_cache_dir(rig: CloudProviderTestRig, tmpdir)
         os.environ["CLOUDPATHLIB_LOCAL_CACHE_DIR"] = original_env_setting
 
 
+def test_environment_variables_force_overwrite(rig: CloudProviderTestRig, tmpdir):
+    # environment instantiation
+    original_env_setting = os.environ.get("CLOUDPATHLIB_FORCE_OVERWRITE_FROM_CLOUD", "")
+
+    try:
```
jayqi, line 351: Is there any reason this is a `try`-`finally` instead of pytest's `monkeypatch.setenv`? https://docs.pytest.org/en/latest/how-to/monkeypatch.html

cloudpathlib/cloudpath.py
```diff
@@ -549,7 +549,7 @@ def open(
         encoding: Optional[str] = None,
         errors: Optional[str] = None,
         newline: Optional[str] = None,
-        force_overwrite_from_cloud: bool = False,  # extra kwarg not in pathlib
+        force_overwrite_from_cloud: Optional[bool] = None,  # extra kwarg not in pathlib
         force_overwrite_to_cloud: bool = False,  # extra kwarg not in pathlib
```
jayqi, line 553: It seems like there are a lot of places where we're not setting default `force_overwrite_to_cloud` to `None`, so `CLOUDPATHLIB_FORCE_OVERWRITE_TO_CLOUD` is not going to get respected.

jayqi (changes requested):

tests/test_caching.py
pjbull: Just to follow the pattern of other tests in this part of the suite. I'm going to leave for now, but if we see any issues, we should switch over at some point.

PR #447, opened 2024-07-02
perf(az): optimize non-recursive `_list_dir`

### LINKED ISSUE #444: Incorrect detection of directories in `AzureBlobPath`
Currently, the detection of whether a given path points to a blob (file) or a directory is based mostly on the presence of its metadata:
https://github.com/drivendataorg/cloudpathlib/blob/08b018b36f90f89003e0c9e5ebd19030b41c2433/cloudpathlib/azure/azblobclient.py#L152-L172

This is not correct, as directories can have non-empty metadata present (and in newer versions of Azure Blob Storage API they almost always do), which leads to incorrect `is_file` and `is_dir` results.

### LINKED ISSUE #446: `AzureBlobPath.iterdir()` is slow for directories with a high level of nesting
`AzureBlobPath.iterdir()` is terribly slow for directories with a high level of nesting, even if most of the blobs are not returned.

This is due to how the current implementation handles non-recursive iteration over a directory, needing to filter *all* of the nested blobs returned by the `list_blobs` function on the `BlobServiceClient`:
https://github.com/drivendataorg/cloudpathlib/blob/08b018b36f90f89003e0c9e5ebd19030b41c2433/cloudpathlib/azure/azblobclient.py#L181-L226

There is another function on the client - `walk_blobs`, which has a parameter allowing for non-recursive iteration over a directory, which might provide a significant performance uplift.

Furthermore, the current implementation does not allow for recursive listing of *containers*, which might be a bit unexpected and currently throws a runtime error.

M0dEx:
As described in #446, the current implementation of `_list_dir` in the `AzureBlobClient` can be painfully slow for directories with a high level of nesting, even if the result is supposed to be non-recursive.

This PR significantly improves the performance by using a newer portion of the Azure Blob Storage API, more specifically the `BlobServiceClient.walk_blobs` function, which allows for much faster non-recursive iteration over the contents of a directory. Additionally, it is now possible to recursively iterate over the root of an account - over all containers.

Furthermore, the detection of whether a blob is a file or a directory has been improved, using `content_type` and `content_md5` properties as described by the [Azure Blob API](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=microsoft-entra-id#response-headers).

Closes #446.
Closes #444.

------------

Performance before:
```
                                  Performance suite results: (2024-07-02T23:25:47.219870)                                  
┏━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ Test Name      ┃ Config Name                ┃ Iterations ┃           Mean ┃              Std ┃            Max ┃ N Items ┃
┡━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│ List Folders   │ List shallow recursive     │         10 │ 0:00:01.153302 │ ± 0:00:00.058301 │ 0:00:01.305643 │   5,500 │
│ List Folders   │ List shallow non-recursive │         10 │ 0:00:01.131045 │ ± 0:00:00.019068 │ 0:00:01.168278 │   5,500 │
│ List Folders   │ List normal recursive      │         10 │ 0:00:01.549334 │ ± 0:00:00.016066 │ 0:00:01.581447 │   7,877 │
│ List Folders   │ List normal non-recursive  │         10 │ 0:00:01.463014 │ ± 0:00:00.022756 │ 0:00:01.503776 │     113 │
│ List Folders   │ List deep recursive        │         10 │ 0:00:02.002217 │ ± 0:00:00.032541 │ 0:00:02.059980 │   7,955 │
│ List Folders   │ List deep non-recursive    │         10 │ 0:00:02.002843 │ ± 0:00:00.021002 │ 0:00:02.038868 │      31 │
│ Glob scenarios │ Glob shallow recursive     │         10 │ 0:00:01.233810 │ ± 0:00:00.021087 │ 0:00:01.263742 │   5,500 │
│ Glob scenarios │ Glob shallow non-recursive │         10 │ 0:00:01.223638 │ ± 0:00:00.022161 │ 0:00:01.255687 │   5,500 │
│ Glob scenarios │ Glob normal recursive      │         10 │ 0:00:01.693786 │ ± 0:00:00.017944 │ 0:00:01.719750 │   7,272 │
│ Glob scenarios │ Glob normal non-recursive  │         10 │ 0:00:01.482890 │ ± 0:00:00.013313 │ 0:00:01.495936 │      12 │
│ Glob scenarios │ Glob deep recursive        │         10 │ 0:00:02.246799 │ ± 0:00:00.018432 │ 0:00:02.277186 │   7,650 │
│ Glob scenarios │ Glob deep non-recursive    │         10 │ 0:00:02.019131 │ ± 0:00:00.018071 │ 0:00:02.046828 │      25 │
│ Walk scenarios │ Walk shallow               │         10 │ 0:00:01.130795 │ ± 0:00:00.022412 │ 0:00:01.171678 │   5,500 │
│ Walk scenarios │ Walk normal                │         10 │ 0:00:01.627434 │ ± 0:00:00.018168 │ 0:00:01.650090 │   7,272 │
│ Walk scenarios │ Walk deep                  │         10 │ 0:00:02.088680 │ ± 0:00:00.044334 │ 0:00:02.156224 │   7,650 │
└────────────────┴────────────────────────────┴────────────┴────────────────┴──────────────────┴────────────────┴─────────┘
```
Performance after:
```
                                  Performance suite results: (2024-07-03T09:40:35.151549)                                  
┏━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ Test Name      ┃ Config Name                ┃ Iterations ┃           Mean ┃              Std ┃            Max ┃ N Items ┃
┡━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│ List Folders   │ List shallow recursive     │         10 │ 0:00:01.148837 │ ± 0:00:00.034477 │ 0:00:01.218684 │   5,500 │
│ List Folders   │ List shallow non-recursive │         10 │ 0:00:01.212517 │ ± 0:00:00.033842 │ 0:00:01.276382 │   5,500 │
│ List Folders   │ List normal recursive      │         10 │ 0:00:01.532350 │ ± 0:00:00.017394 │ 0:00:01.550038 │   7,272 │
│ List Folders   │ List normal non-recursive  │         10 │ 0:00:00.023452 │ ± 0:00:00.009190 │ 0:00:00.049427 │     113 │
│ List Folders   │ List deep recursive        │         10 │ 0:00:01.707918 │ ± 0:00:00.036464 │ 0:00:01.782752 │   7,650 │
│ List Folders   │ List deep non-recursive    │         10 │ 0:00:00.028419 │ ± 0:00:00.004629 │ 0:00:00.039337 │      31 │
│ Glob scenarios │ Glob shallow recursive     │         10 │ 0:00:01.261963 │ ± 0:00:00.034387 │ 0:00:01.343164 │   5,500 │
│ Glob scenarios │ Glob shallow non-recursive │         10 │ 0:00:01.299316 │ ± 0:00:00.024798 │ 0:00:01.353754 │   5,500 │
│ Glob scenarios │ Glob normal recursive      │         10 │ 0:00:01.677940 │ ± 0:00:00.021779 │ 0:00:01.717995 │   7,272 │
│ Glob scenarios │ Glob normal non-recursive  │         10 │ 0:00:00.022214 │ ± 0:00:00.000882 │ 0:00:00.023246 │      12 │
│ Glob scenarios │ Glob deep recursive        │         10 │ 0:00:01.900239 │ ± 0:00:00.023626 │ 0:00:01.947988 │   7,650 │
│ Glob scenarios │ Glob deep non-recursive    │         10 │ 0:00:00.026393 │ ± 0:00:00.000928 │ 0:00:00.028201 │      25 │
│ Walk scenarios │ Walk shallow               │         10 │ 0:00:01.148723 │ ± 0:00:00.019138 │ 0:00:01.190942 │   5,500 │
│ Walk scenarios │ Walk normal                │         10 │ 0:00:01.555491 │ ± 0:00:00.016058 │ 0:00:01.580223 │   7,272 │
│ Walk scenarios │ Walk deep                  │         10 │ 0:00:01.707682 │ ± 0:00:00.017401 │ 0:00:01.733078 │   7,650 │
└────────────────┴────────────────────────────┴────────────┴────────────────┴──────────────────┴────────────────┴─────────┘
```

### DIFF
--- a/cloudpathlib/azure/azblobclient.py
+++ b/cloudpathlib/azure/azblobclient.py
@@ -1,7 +1,7 @@
 from datetime import datetime, timedelta
 import mimetypes
 import os
-from pathlib import Path, PurePosixPath
+from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union
 
 
@@ -20,6 +20,7 @@ try:
         BlobProperties,
         ContentSettings,
         generate_blob_sas,
+        BlobPrefix,
     )
 except ModuleNotFoundError:
     implementation_registry["azure"].dependencies_loaded = False
@@ -181,17 +182,14 @@ class AzureBlobClient(Client):
     def _list_dir(
         self, cloud_path: AzureBlobPath, recursive: bool = False
     ) -> Iterable[Tuple[AzureBlobPath, bool]]:
-        # shortcut if listing all available containers
         if not cloud_path.container:
-            if recursive:
-                raise NotImplementedError(
-                    "Cannot recursively list all containers and contents; you can get all the containers then recursively list each separately."
-                )
+            for container in self.service_client.list_containers():
+                yield self.CloudPath(f"az://{container.name}"), True
 
-            yield from (
-                (self.CloudPath(f"az://{c.name}"), True)
-                for c in self.service_client.list_containers()
-            )
+                if not recursive:
+                    continue
+
+                yield from self._list_dir(self.CloudPath(f"az://{container.name}"), recursive=True)
             return
 
         container_client = self.service_client.get_container_client(cloud_path.container)
@@ -200,30 +198,30 @@ class AzureBlobClient(Client):
         if prefix and not prefix.endswith("/"):
             prefix += "/"
 
-        yielded_dirs = set()
+        if not recursive:
+            blobs = container_client.walk_blobs(name_starts_with=prefix)
+        else:
+            blobs = container_client.list_blobs(name_starts_with=prefix)
 
-        # NOTE: Not recursive may be slower than necessary since it just filters
-        #   the recursive implementation
-        for o in container_client.list_blobs(name_starts_with=prefix):
-            # get directory from this path
-            for parent in PurePosixPath(o.name[len(prefix) :]).parents:
-                # if we haven't surfaced this directory already
-                if parent not in yielded_dirs and str(parent) != ".":
-                    # skip if not recursive and this is beyond our depth
-                    if not recursive and "/" in str(parent):
-                        continue
+        def is_dir(blob: Union[BlobProperties, BlobPrefix]) -> bool:
+            # walk_blobs returns a BlobPrefix object for directories
+            # https://learn.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.blobprefix?view=azure-python
+            if isinstance(blob, BlobPrefix):
+                return True
 
-                    yield (
-                        self.CloudPath(f"az://{cloud_path.container}/{prefix}{parent}"),
-                        True,  # is a directory
-                    )
-                    yielded_dirs.add(parent)
+            # content_type and content_md5 are never None for files in Azure Blob Storage
+            # if they are None, then the given path is a directory
+            # https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=microsoft-entra-id#response-headers
+            return (
+                blob.content_settings.content_md5 is None
+                and blob.content_settings.content_type is None
+            )
 
-            # skip file if not recursive and this is beyond our depth
-            if not recursive and "/" in o.name[len(prefix) :]:
-                continue
-
-            yield (self.CloudPath(f"az://{cloud_path.container}/{o.name}"), False)  # is a file
+        for blob in blobs:
+            # walk_blobs returns folders with a trailing slash
+            blob_path = blob.name.rstrip("/")
+            blob_cloud_path = self.CloudPath(f"az://{cloud_path.container}/{blob_path}")
+            yield blob_cloud_path, is_dir(blob)
 
     def _move_file(
         self, src: AzureBlobPath, dst: AzureBlobPath, remove_src: bool = True
--- a/cloudpathlib/azure/azblobclient.py
+++ b/cloudpathlib/azure/azblobclient.py
@@ -2,8 +2,7 @@ from datetime import datetime, timedelta
 import mimetypes
 import os
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union
-
+from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union, cast
 
 from ..client import Client, register_client_class
 from ..cloudpath import implementation_registry
@@ -156,7 +155,20 @@ class AzureBlobClient(Client):
             return "dir"
 
         try:
-            self._get_metadata(cloud_path)
+            # the result of get_blob_properties is a BlobProperties object (dict mixin)
+            metadata = cast(BlobProperties, self._get_metadata(cloud_path))
+
+            # content_type and content_md5 are never None for files in Azure Blob Storage
+            # if they are None, then the given path is a directory
+            # https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=microsoft-entra-id#response-headers
+            is_folder = (
+                metadata.content_settings.content_type is None
+                and metadata.content_settings.content_md5 is None
+            )
+
+            if is_folder:
+                return "dir"
+
             return "file"
         except ResourceNotFoundError:
             prefix = cloud_path.blob
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -8,6 +8,8 @@
 - fix: use native `exists()` method in `GSClient`. (PR [#420](https://github.com/drivendataorg/cloudpathlib/pull/420))
 - Enhancement: lazy instantiation of default client (PR [#432](https://github.com/drivendataorg/cloudpathlib/issues/432), Issue [#428](https://github.com/drivendataorg/cloudpathlib/issues/428))
 - Adds existence check before downloading in `download_to` (Issue [#430](https://github.com/drivendataorg/cloudpathlib/issues/430), PR [#432](https://github.com/drivendataorg/cloudpathlib/pull/432))
+- perf(az): optimize non-recursive _list_dir (Issue [#446](https://github.com/drivendataorg/cloudpathlib/issues/446), PR [#447](https://github.com/drivendataorg/cloudpathlib/pull/447))
+- fix(az): improve detection of path types (Issue [#444](https://github.com/drivendataorg/cloudpathlib/issues/444), PR [#447](https://github.com/drivendataorg/cloudpathlib/pull/447))
 
 ## v0.18.1 (2024-02-26)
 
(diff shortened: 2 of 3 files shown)


pjbull:
Thanks, @M0dEx. Could you also run before and after perf numbers for Azure?

You'll need to run this command but with `az` instead of `s3`:
https://github.com/drivendataorg/cloudpathlib/blob/master/Makefile#L89-L90

You should get a table you can paste into the PR like this:
https://cloudpathlib.drivendata.org/stable/contributing/#performance-testing

M0dEx:
> Thanks, @M0dEx. Could you also run before and after perf numbers for Azure?
> 
> You'll need to run this command but with `az` instead of `s3`: https://github.com/drivendataorg/cloudpathlib/blob/master/Makefile#L89-L90
> 
> You should get a table you can paste into the PR like this: https://cloudpathlib.drivendata.org/stable/contributing/#performance-testing

Added performance tests before and after.

jayqi:
Looks like the performance generally is the same, with `normal` and `deep` cases being much faster for `List` and `Glob` as advertised. 👍

M0dEx:
Pushed a fix for the failing tests (apparently caused by the `|=` operator not working as expected in Python 3.8).

M0dEx:
Removed the `|` union as well, since it is only supported since Python 3.10.

cloudpathlib/azure/azblobclient.py
```diff
@@ -200,30 +210,30 @@ def _list_dir(
...
-                    )
-                    yielded_dirs.add(parent)
-
-            # skip file if not recursive and this is beyond our depth
-            if not recursive and "/" in o.name[len(prefix) :]:
-                continue
-
-            yield (self.CloudPath(f"az://{cloud_path.container}/{o.name}"), False)  # is a file
+        if not recursive:
+            blobs = container_client.walk_blobs(name_starts_with=prefix)
```
pjbull, line 214: Does this introduce an alternative minimum sdk version to what we have?
https://github.com/drivendataorg/cloudpathlib/blob/master/pyproject.toml#L36

cloudpathlib/azure/azblobclient.py
```diff
@@ -155,7 +155,20 @@ def _is_file_or_dir(self, cloud_path: AzureBlobPath) -> Optional[str]:
...
+            # the result of get_blob_properties is a BlobProperties object (dict mixin)
+            metadata = cast(BlobProperties, self._get_metadata(cloud_path))
+
+            # content_type and content_md5 are never None for files in Azure Blob Storage
+            # if they are None, then the given path is a directory
+            # https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=microsoft-entra-id#response-headers
+            is_folder = (
+                metadata.content_settings.content_type is None
+                and metadata.content_settings.content_md5 is None
+            )
```
pjbull, line 167: Can you point to a more specific reference? I don't see the link you sent supporting the claim that this is definitive for folders. 

It does seem like `x-ms-resource-type` is useful to answer the question directly, but only for certain configurations so could be included as an optimization. It may be the case that all the scenarios where those vars are none are also ones where `x-ms-resource-type` is available.

cloudpathlib/azure/azblobclient.py
```diff
@@ -200,30 +210,30 @@ def _list_dir(
...
+            # if they are None, then the given path is a directory
+            # https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=microsoft-entra-id#response-headers
+            return (
+                blob.content_settings.content_md5 is None
+                and blob.content_settings.content_type is None
+            )
+
+        for blob in blobs:
+            # walk_blobs returns folders with a trailing slash
+            blob_path = blob.name.rstrip("/")
```
pjbull, line 234: Is it problematic to keep the trailing slashes?

pjbull (changes requested):
I've got a few questions about the approach here. Preferably we can use a more explicitly supported version of things.

cloudpathlib/azure/azblobclient.py
M0dEx: I did it to be consistent with the current behaviour, but we can simply not remove the trailing slashes.

cloudpathlib/azure/azblobclient.py
M0dEx: No, it does not:
https://github.com/Azure/azure-sdk-for-python/blob/0ab186f98cf2c4c83c1b4260406027d9c4f8cc06/sdk/storage/azure-storage-blob/azure/storage/blob/_container_client.py#L652

cloudpathlib/azure/azblobclient.py
M0dEx: I agree the reference is sort of vague. I was mainly going of off these two statements:
> Content-Type: The content type that's specified for the blob. If no content type is specified, the default content type is application/octet-stream.

> Content-MD5: If the Content-MD5 header has been set for the blob, this response header is returned so that the client can check for message content integrity.
> In version 2012-02-12 and later, Put Blob sets a block blob’s MD5 value even when the Put Blob request doesn’t include an MD5 header.

Using this, coupled with our observations from real-life usage, we concluded that these parameters are always `None` for folders and never `None` for blobs, even if they are empty.

pjbull: Does your configuration have the `x-ms-resource-type` header for these folder blobs?

M0dEx: That header does not seem to be returned within `BlobProperties`, nor can I see it in the in-browser Azure storage explorer.

pjbull:
Incorporated into #453, thanks!

PR #449, opened 2024-07-02
fix(az): correctly remove nested file structures

### LINKED ISSUE #448: `AzureBlobPath.rmtree()` sometimes fails and needs to be called again
In certain scenarios, `AzureBlobPath.rmtree()` can fail and needs to be called again to finish removing the nested file structure. Even after some investigation, we have been unable to pinpoint and reproduce the issue reliably. We suspect the reason might be some sort of a race condition on the side of the Azure Blob Storage API.

M0dEx:
As described in #448, the current `AzureBlobPath.rmtree()` can sometimes fail and must be called again to remove *all* of the file/directory structure.

This PR fixes this issue by first deleting all blobs, then deleting all folders, from the most nested to the least nested ones, and finally removing the root folder. This guarantees that the operation will always succeed.

Closes #448.

### DIFF
--- a/cloudpathlib/azure/azblobclient.py
+++ b/cloudpathlib/azure/azblobclient.py
@@ -250,24 +250,32 @@ class AzureBlobClient(Client):
 
         return dst
 
-    def _remove(self, cloud_path: AzureBlobPath, missing_ok: bool = True) -> None:
+    def _remove(self, cloud_path: AzureBlobPath, missing_ok: bool = True) -> None:  # type: ignore
+        container_client = self.service_client.get_container_client(cloud_path.container)
         file_or_dir = self._is_file_or_dir(cloud_path)
-        if file_or_dir == "dir":
-            blobs = [
-                b.blob for b, is_dir in self._list_dir(cloud_path, recursive=True) if not is_dir
-            ]
-            container_client = self.service_client.get_container_client(cloud_path.container)
-            container_client.delete_blobs(*blobs)
-        elif file_or_dir == "file":
-            blob = self.service_client.get_blob_client(
-                container=cloud_path.container, blob=cloud_path.blob
-            )
 
-            blob.delete_blob()
-        else:
-            # Does not exist
-            if not missing_ok:
-                raise FileNotFoundError(f"File does not exist: {cloud_path}")
+        if not file_or_dir:
+            if missing_ok:
+                return
+
+            raise FileNotFoundError(f"File does not exist: {cloud_path}")
+
+        if file_or_dir == "dir":
+            blobs = [(blob, is_dir) for blob, is_dir in self._list_dir(cloud_path, recursive=True)]
+
+            # need to delete files first to allow deleting the folders
+            files = [blob.blob for blob, is_dir in blobs if not is_dir]
+            container_client.delete_blobs(*files)
+
+            # folders need to be deleted from the deepest to the shallowest
+            folders = sorted(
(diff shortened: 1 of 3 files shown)


pjbull:
@M0dEx generally this looks good on initial review. Is there a test case you can add that fails against the live server without this fix and succeeds with it?

M0dEx:
> @M0dEx generally this looks good on initial review. Is there a test case you can add that fails against the live server without this fix and succeeds with it?

Since the root cause might be specific container-level settings (retention, deletion policies, etc.), it has been difficult for us to replicate the issue reliably, especially using Azurite, which only provides a subset of the Azure Blob API features.

I might be asking you to take a lot of faith here, but this change has fixed the behaviour of `rmtree()` in the mentioned use cases for us.

cloudpathlib/azure/azblobclient.py
```diff
@@ -250,24 +250,32 @@ def _move_file(
...
-                container=cloud_path.container, blob=cloud_path.blob
+            blobs = [(blob, is_dir) for blob, is_dir in self._list_dir(cloud_path, recursive=True)]
+
+            # need to delete files first to allow deleting the folders
+            files = [blob.blob for blob, is_dir in blobs if not is_dir]
+            container_client.delete_blobs(*files)
+
+            # folders need to be deleted from the deepest to the shallowest
+            folders = sorted(
+                (blob.blob for blob, is_dir in blobs if is_dir and blob.exists()), reverse=True
```
pjbull, line 272: I'm a little worried about perf since `.exists` adds a network call. This means that for large file trees (broad or deep) we potentially add a lot of calls for "fake" folders that don't actually exist on storage but `_list_dir` returns to act like a file system. 

I believe that only explicitly created folders on blob storage (or maybe with certain parameters set) would still stick around without specific removal. To create these, you need to be using hierarchical namespaces/a Data Lake Gen2 Storage Account.

I think the fix may instead to be to use an azure-SDK API that will list all explicit blobs (files or folders) rather than `_list_dir` in determining what to remove. I'm not sure if `list_blobs` does that under accounts with the hierarchical namespaces, so I think we'd need to test that first to see if it works for your use cases.

Do you think you could dig in and see if those settings/accounts can repro the issue for you?

pjbull (changes requested):
I think the network overhead is not worth it for most Azure configs. See inline note.

cloudpathlib/azure/azblobclient.py
M0dEx: > I'm a little worried about perf since .exists adds a network call. This means that for large file trees (broad or deep) we potentially add a lot of calls for "fake" folders that don't actually exist on storage but _list_dir returns to act like a file system.

I don't think the call to `exists` here is strictly necessary, but I might be misremembering.

> I believe that only explicitly created folders on blob storage (or maybe with certain parameters set) would still stick around without specific removal. To create these, you need to be using hierarchical namespaces/a Data Lake Gen2 Storage Account.

Hmm, on further investigation, it seems the "problematic" accounts are using hierarchical namespaces. The directories seems to be "sticky" even without being explicitly created, which is a bit surprising.

pjbull: That's very helpful to know, thanks. I'd love to support that feature of blob storage—I'll set one up on our test account and see how it works. We may need more fixes than this to make `cloudpathlib` play nice with hierarchical namespace enabled storage accounts.

pjbull:
Incorporated into #453, thanks!

PR #453, opened 2024-07-17
Initial ADLS gen2 support

### LINKED ISSUE #161: Support Azure Data Lake Storage Gen2
Not sure if this problem also affects regular Blob storage accounts, but I have a Azure storage account using Data Lake Gen2 storage, and the I observe the following issues:

* All paths are considered to be files (`is_file()` is true if it exists) and not directories. `is_dir()` never returns True
* `iterdir()` always returns paths. If the input is a file, it just returns itself if it exists.
* `rmdir()` always throws a `CloudPathNotADirectoryError`
* `unlink()` does appear to remove empty directories, but is expected to throw `CloudPathIsADirectoryError`
* `rmtree()` throws a `CloudPathNotADirectoryError`
* `rename()` (if it works) copies contents instead of doing an actual rename, which is supported in Gen2 (#162)

There may be other issues.

pjbull:
Add initial ADLS gen2 support:

 - Accept or create a `data_lake_service_client`
 - When necessary in `AzureBlobClient` use the `data_lake_service_client` API. This does _not_ use the ADLS API everywhere possible since ADLS Gen2 works with the normal blob API in many places. Do this since it is common to enable hierarchical namespaces when creating a blob storage account, not realizing that it needs ADLS APIs for some scenarios.
 - Add the ADLS SDK as a dependency
 - Adds a new Azure test rig for Gen2 if configured
 - Configures ADLS Gen2 testing on Github Actions

This also incorporates the fixes in:
 - PR #447 
 - PR #449

Closes #161

----------

 - [x] Test the different instantiation pipelines
 - [x] Test empty folders on ADLS Gen2
 - [x] Fix mocks to do offline tests for ADLS Gen2 and not
 - [x] Add additionally functionality mentioned in #161


cloudpathlib/azure/azblobclient.py
```diff
@@ -200,30 +260,24 @@ def _list_dir(
...
+                blobs = container_client.walk_blobs(name_starts_with=prefix)
+            else:
+                blobs = container_client.list_blobs(name_starts_with=prefix)
 
-            yield (self.CloudPath(f"az://{cloud_path.container}/{o.name}"), False)  # is a file
+            for blob in blobs:
+                # walk_blobs returns folders with a trailing slash
+                blob_path = blob.name.rstrip("/")
+                blob_cloud_path = self.CloudPath(f"az://{cloud_path.container}/{blob_path}")
+                yield blob_cloud_path, isinstance(blob, BlobPrefix)
```
M0dEx, line 280: Can you rely on `isinstance(blob, BlobPrefix)` to determine whether the blob is a directory even in the recursive case (`list_blobs()`)?

pjbull: Short answer is yes (there are no folders with `list_blobs` if we are not on an HNS storage account, and the HNS accounts are handled above by the ADLS API). I updated to make more explicit.

cloudpathlib/azure/azblobclient.py
```diff
@@ -94,27 +104,101 @@ def __init__(
...
 
-    def _get_metadata(self, cloud_path: AzureBlobPath) -> Union["BlobProperties", Dict[str, Any]]:
-        blob = self.service_client.get_blob_client(
-            container=cloud_path.container, blob=cloud_path.blob
-        )
-        properties = blob.get_blob_properties()
+        self._hns_enabled = None
+
+    def _check_hns(self) -> Optional[bool]:
+        if not self._hns_enabled:
```
pjbull, line 173: Should be an `is None` check

cloudpathlib/azure/azblobclient.py
```diff
@@ -238,6 +327,16 @@ def _move_file(
                 metadata=dict(last_modified=str(datetime.utcnow().timestamp()))
             )
 
+        # we can use rename API same account and container on adls gen2
```
pjbull, line 330: Update comment; does not have to be same container

docs/docs/authentication.md
```diff
@@ -211,6 +211,13 @@ client.set_as_default_client()
 cp3 = CloudPath("s3://cloudpathlib-test-bucket/")
 ```
 
+## Accessing Azure DataLake Storage Gen2 (ADLS Gen2) storage with hierarchical namespace enabled
+
+Some Azure storage accounts are configured with "hierarchical namespace" enabled. This means that the storage account is backed by the Azure DataLake Storage Gen2 product rather than Azure Blob Storage. For many operations, the two are the same and one can use the Azure Blob Storage API. However, for some operations, a developer will need to use the Azure DataLake Storage API. The `AzureBlobClient` class implemented in cloudpathlib is designed to detect if hierarchical namespace is enabled and use the Azure DataLake Storage API in the places where it is necessary or it provides a performance improvement. Usually, a user of cloudpathlib will not need to know if hierarchical namespace is enabled and the storage account is backed by Azure DataLake Storage Gen2 or Azure Blob Storage.
+
+If needed, the Azure SDK provided `DataLakeServiceClient` object can be accessed via the `AzureBlobClient.data_lake_client`. The Azure SDK provided `BlobServiceClient` object can be accessed via `AzureBlobClient.blob_client`.
```
pjbull, line 218: should be `AzureBlobClient.service_client`

tests/mock_clients/mock_azureblob.py
```diff
@@ -17,51 +17,82 @@
...
-        @property
-        def credential(self):
-            """Returns well-known account key used by Azurite
-            See: https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio%2Cblob-storage#well-known-storage-account-and-key
-            """
-            return SharedKeyCredentialPolicy(
-                self.account_name,
-                "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==",
-            )
+class _JsonCache:
```
pjbull, line 20: Add comment for what this is used for

tests/mock_clients/mock_azureblob.py
```diff
@@ -17,51 +17,82 @@
...
-            return MockContainerClient(self.tmp_path, container_name=container)
+class MockBlobServiceClient:
+    def __init__(self, test_dir, adls):
+        # copy test assets for reference in tests without affecting assets
+        shutil.copytree(TEST_ASSETS, test_dir, dirs_exist_ok=True)
 
-        def list_containers(self):
-            Container = namedtuple("Container", "name")
-            return [Container(name=DEFAULT_CONTAINER_NAME)]
+        # root is parent of the test specific directort
```
pjbull, line 52: Fix typo

pjbull (commented):
A couple of typos/tweaks

pjbull:
@M0dEx Thanks for your patience—ended up being more here than I expected. If you've got time, it would be great if you took this branch for a test drive and took a look at the code changes.

pjbull:
Planning to rebase on top of https://github.com/drivendataorg/cloudpathlib/pull/461 when that is in so that tests are reliable.

pjbull:
@jayqi Should be ready for the release. Scope/maintenance burden is a little larger, but I think this is a large standing issue with a fair amount of interest so it is worth it to get in.

jayqi (approved):
Okay, it feels a little messy but I think it's reasonable.

pjbull:
> Okay, it feels a little messy but I think it's reasonable.

Truer words were never spoken.

Issue #450, opened 2024-07-03
Multiple pickle roundtrip serializations cause KeyError


kujenga:
The following code causes a crash, which can arise when passing cloudpathlib objects to/from subprocesses.

```python
    path1 = CloudPath("s3://bucket/key")
    print(path1.__dict__)
    pkl1 = pickle.dumps(path1)

    path2 = pickle.loads(pkl1)
    print(path2.__dict__)
    pkl2 = pickle.dumps(path2)

    path3 = pickle.loads(pkl2)
    print(path3.__dict__)
```

The exception raised is:

```
{'_handle': None, '_client': None, '_str': 's3://bucket/key', '_url': ParseResult(scheme='s3', netloc='bucket', path='/key', params='', query='', fragment=''), '_path': PurePosixPath('/bucket/key'), '_dirty': False}
{'_handle': None, '_str': 's3://bucket/key', '_url': ParseResult(scheme='s3', netloc='bucket', path='/key', params='', query='', fragment=''), '_path': PurePosixPath('/bucket/key'), '_dirty': False}
Traceback (most recent call last):
  File "/path/to/cloudpathlib-pickle-crash.py", line 20, in <module>
    main()
  File "/path/to/cloudpathlib-pickle-crash.py", line 13, in main
    pkl2 = pickle.dumps(path2)
           ^^^^^^^^^^^^^^^^^^^
  File "/path/to/.venv/lib/python3.11/site-packages/cloudpathlib/cloudpath.py", line 266, in __getstate__
    del state["_client"]
        ~~~~~^^^^^^^^^^^
KeyError: '_client'
```

The issue seems to be that these methods are not symmetrical, and the _client field is not restored when the object is unpickled in a way that might mirror this example in the docs https://docs.python.org/3/library/pickle.html#pickle-state at this point in the code: https://github.com/drivendataorg/cloudpathlib/blob/08b018b36f90f89003e0c9e5ebd19030b41c2433/cloudpathlib/cloudpath.py#L262-L272

pjbull:
Thanks @kujenga.

Since [`_client` is created on demand](https://github.com/drivendataorg/cloudpathlib/blob/08b018b36f90f89003e0c9e5ebd19030b41c2433/cloudpathlib/cloudpath.py#L245-L250) I think the fix here is just to update `__getstate__` with:

```python
if "_client" in state:
    del state["_client"]
```

Would be happy to take a PR that does that and adds a test.

Thanks!

[...]

PR #454, opened 2024-07-21
allow for roundtrip of cloudpaths through pickle serialization

kujenga:
This avoids an exception thrown because the client is not serialized into the pickeld object, and thus when __getstate__ is called the second time, there is no _client field to delete.

Closes https://github.com/drivendataorg/cloudpathlib/issues/450

----------------

Contributor checklist:

 - [ ] Tested manually against live server backend for at least one provider
 - [x] Linting passes locally
 - [x] Tests pass locally
 - [ ] Updated `HISTORY.md` with the issue that is addressed and the PR you are submitting. If the top section is not `## UNRELEASED``, then you need to add a new section to the top of the document for your change.


tests/test_cloudpath_serialize.py
```diff
@@ -0,0 +1,14 @@
+import pickle
+
+from cloudpathlib import CloudPath
+
+
+def test_pickle_roundtrip():
```
kujenga, line 6: If there is an existing file this would be better suited to put it within let me know and happy to move it!

I verified that this does repro the issue without the above change.

pjbull: Separate file for serializing is fine. Can you also move this test over to the new file so they are together? https://github.com/drivendataorg/cloudpathlib/blob/08b018b36f90f89003e0c9e5ebd19030b41c2433/tests/test_cloudpath_file_io.py#L457-L472

pjbull (changes requested):

pjbull:
Looks good, thanks! One small ask to move the other pickle test to the same place

pjbull:
@kujenga Planning a release soon—will you have a chance to wrap this PR?

kujenga:
Thank you very much! Apologies I missed the earlier notification

PR #462, opened 2024-08-22
Change LocalClient to not explicitly store default storage directory

jayqi:
This is my proposed alternative fix for #414. This PR is into the branch for #436.

This changes the way that `LocalClient` stores and accesses what the local storage directory should be. In the default case where no value for `local_storage_dir` is provided, accessing local storage will call the class method `get_default_storage_dir` to locate local storage. 

This means that calling `reset_default_storage_dir` will instantly reset the local storage for all clients that are using the default storage, not just future new ones that are created, and not with a special case for the default client. 

This feels like a more intuitive fix to me than having the special case for the default client. This means that calling `reset_default_storage_dir` will have the same effect as deleting all of the files inside the directory, in the sense that all instantiated clients using that storage directory get affected.

### DIFF
--- a/test.py
+++ b/test.py
@@ -1,7 +1,0 @@
-from cloudpathlib import S3Path, S3Client
-
-client = S3Client(profile_name="bad_profile")
-
-import gc
-
-gc.collect()
--- a/test.py
+++ b/test.py
@@ -1,0 +1,7 @@
+from cloudpathlib import S3Path, S3Client
+
+client = S3Client(profile_name="bad_profile")
+
+import gc
+
+gc.collect()
--- a/tests/test_local.py
+++ b/tests/test_local.py
@@ -59,6 +59,8 @@ def test_default_storage_dir(client_class, monkeypatch):
     p1.write_text("hello")
     assert p1.exists()
     assert p1.read_text() == "hello"
+
+    # p2 uses a new client, but the simulated "cloud" should be the same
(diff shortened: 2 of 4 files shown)


tests/test_local.py
```diff
@@ -76,20 +78,30 @@ def test_reset_default_storage_dir(client_class, monkeypatch):
     cloud_prefix = client_class._cloud_meta.path_class.cloud_prefix
 
     p1 = client_class().CloudPath(f"{cloud_prefix}drive/file.txt")
-    client_class.reset_default_storage_dir()
-    p2 = client_class().CloudPath(f"{cloud_prefix}drive/file.txt")
-
     assert not p1.exists()
-    assert not p2.exists()
```
jayqi, line 83: This test fails with my new implementation, but I think the old behavior is not desirable. In the old version of this test, we reset the storage directory before instantiating `p2` and writing out `p1`, which means `p2` uses a new default local storage but `p1` still uses the old default local storage.

test.py
```diff
@@ -0,0 +1,7 @@
+from cloudpathlib import S3Path, S3Client
+
+client = S3Client(profile_name="bad_profile")
+
+import gc
+
+gc.collect()
```
pjbull, line 7: Assume this is leftover?

pjbull (approved):
Looks great, thanks!

PR #464, opened 2024-08-23
feat: Add overloads for CloudPath.open()

### LINKED ISSUE #465: Add typing overloads to `CloudPath.open()`
Currently, the return type of `CloudPath.open()` is always `IO[Any]`. This differs from `pathlib.Path.open()`, which returns more specific types like `TextIOWrapper`, `FileIO`, `BufferedRandom`, `BufferedWriter`, `BufferedReader`, `BinaryIO`, or `IO[Any]` depending on arguments such as `mode`, `buffering`, and `encoding`. The more specific return types are useful for catching typing issues where a library expects a file handle in binary or text mode, specified as `IO[bytes]` or `IO[str]`, respectively. With the current behavior, passing a file handle returned by `CloudPath.open()` often requires using `typing.cast()` or `# type: ignore`.

Here’s an example illustrating the current behavior compared to `pathlib.Path.open()`:

```python
from pathlib import Path

cloud_path = CloudPath("s3://bucket/key")  # type: ignore
local_path = Path("test.txt")

with local_path.open("w") as f:
    reveal_type(f)  # Revealed type: io.TextIOWrapper

with local_path.open("wb") as f:
    reveal_type(f)  # Revealed type: io.BufferedWriter

with cloud_path.open("w") as f:
    reveal_type(f)  # Revealed type: typing.IO[Any]

with cloud_path.open("wb") as f:
    reveal_type(f)  # Revealed type: typing.IO[Any]
```

0xbe7a:
Closes #465

To correctly handle the different argument and return types, I adapted the `typeshed` overloads for `pathlib.Path.open()` to `CloudPath.open()`. These overloads are appropriate because `cloudpathlib` passes the arguments to `self._local` after caching the file to local disk.

Handling all possible `mode` values is verbose, but the overloads are copied directly from `typeshed`, ensuring correctness and minimizing maintenance burden.

----------------

Contributor checklist:

 - (Not applicable) Tested manually against live server backend for at least one provider
 - (Not applicable) Added tests for any new functionality
 - [x] Linting passes locally
 - [x] Tests pass locally


cloudpathlib/typing.py
```diff
@@ -0,0 +1,71 @@
...
+    "xb+",
+    "x+b",
+    "+xb",
+    "bx+",
+    "b+x",
+    "+bx",
+]
+OpenBinaryModeWriting: TypeAlias = Literal["wb", "bw", "ab", "ba", "xb", "bx"]
+OpenBinaryModeReading: TypeAlias = Literal["rb", "br", "rbU", "rUb", "Urb", "brU", "bUr", "Ubr"]
+OpenBinaryMode: TypeAlias = OpenBinaryModeUpdating | OpenBinaryModeReading | OpenBinaryModeWriting
```
pjbull, line 71: I don't think we want to vendor and maintain these. We should remove this file.

cloudpathlib/cloudpath.py
```diff
@@ -34,6 +37,14 @@
...
 from warnings import warn
 
+from cloudpathlib.typing import (
+    OpenBinaryMode,
+    OpenBinaryModeReading,
+    OpenBinaryModeUpdating,
+    OpenBinaryModeWriting,
+    OpenTextMode,
+)
+
```
pjbull, line 47: Import these directly from `_typeshed` instead of vendoring:

```python
if TYPE_CHECKING:
    from _typeshed import OpenBinaryMode, OpenBinaryModeReading, OpenBinaryModeUpdating, OpenBinaryModeWriting, OpenTextMode
```

You'll need to put the references in quotations since they won't be available at runtime.

pjbull (changes requested):
Thanks! A few changes requested. If you update soon, happy to include this in the next release coming in the next few days.

cloudpathlib/cloudpath.py
0xbe7a: Done!

cloudpathlib/typing.py
0xbe7a: I have dropped the vendored aliases

PR #467, opened 2024-08-31
use variable for client schemes, allowing override

### LINKED ISSUE #466: Allow subclasses of S3Client, etc to set a custom scheme
Currently, the guidance for accessing S3 compatible stores recommends continuing to use the S3 scheme in the path: https://cloudpathlib.drivendata.org/stable/authentication/#accessing-custom-s3-compatible-object-stores

I'd like to be able to take this a step further and customize the URL scheme. The following currently runs into an issue as `cloud_prefix` is not available on the client implementations, which instead hard-code their scheme.

```python
@register_path_class("mys3")
class MyS3Path(S3Path):

    cloud_prefix: str = "mys3://"


@register_client_class("mys3")
class MyS3Client(S3Client):

    cloud_prefix: str = "mys3://"
```

The idea here is to allow the schemes to be customized in a similar way to how the Path implementations work: https://github.com/drivendataorg/cloudpathlib/blob/b776bee3a6cdbb5877cd6eb8be823af0d0456e0c/cloudpathlib/s3/s3path.py#L28

kujenga:
This change is intended to make the default client implementations more flexible so that their scheme can be customized. This can be useful in scenarios where a subclass wants to implement a custom scheme on e.g. a S3 compatible API [1] but with a custom scheme so that the default S3 access is still also available.

[1] https://cloudpathlib.drivendata.org/stable/authentication/#accessing-custom-s3-compatible-object-stores

The tests have been updated to include a new s3-like rig which uses the
new scheme override functionality.

(If there is a better strategy for tests or this seems like overkill let me know and happy to adjust!)

Closes #466

----------------

Contributor checklist:

 - [ ] Added tests for any new functionality
 - [x] Linting passes locally
 - [x] Tests pass locally
 - [ ] Updated `HISTORY.md` with the issue that is addressed and the PR you are submitting. If the top section is not `## UNRELEASED``, then you need to add a new section to the top of the document for your change.


pjbull (changes requested):
Yeah, the hard-coding is a bug. Thanks for the fix. Agreed we want to enable scenarios like you lay out.

Do we need to add `cloud_prefix` to the client though? I'd rather keep it in one place just on the path.

For all the instances you replaced, you should be able to get the `cloud_prefix` from the instance of the `cloud_path` passed to the method, no?

Would that work for your use cases? If not, we could potentially add `cloud_prefix` as a property that reaches through `_cloud_meta` to get the associated `cloud_prefix` from the `path_class`:
https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/cloudpath.py#L96-L100

pjbull:
Also, one other thought: I think adding a rig for this is overkill. That will duplicate all the tests, which isn't necessary. I'd just do a single test with the custom scheme to show that piece works.

kujenga:
> For all the instances you replaced, you should be able to get the cloud_prefix from the instance of the cloud_path passed to the method, no?

Good call, done!

> Also, one other thought: I think adding a rig for this is overkill. That will duplicate all the tests, which isn't necessary. I'd just do a single test with the custom scheme to show that piece works.

Fair enough, done!

If there's anything else just let me know! I left them as separate commits to keep what I changed clear but happy to squash as well if you prefer.

tests/test_client.py
```diff
@@ -116,3 +119,34 @@ def my_content_type(path):
...
+
+
+@register_path_class("mys3")
+class MyS3Path(S3Path):
+    cloud_prefix: str = "mys3://"
+
+
+@register_client_class("mys3")
+class MyS3Client(S3Client):
+    pass
```
pjbull, line 131: Wondering if we can isolate these in case they cause some weird interaction failure happens in the future, e.g., in a fixture:

```python
@pytest.fixture
def custom_s3_path():
    @register_path_class("mys3")
    class MyS3Path(S3Path):
        cloud_prefix: str = "mys3://"
  
    @register_client_class("mys3")
    class MyS3Client(S3Client):
        pass

    yield

    # cleanup after use
    implementation_registry.pop("mys3")
```

pjbull (changes requested):
One last tweak to the tests and could you add a `HISTORY.md` comment? Then I think it's good to go. Thanks!

tests/test_client.py
kujenga: Yep! makes sense.

Typing takes a slight hit within the tests (I tried `custom_s3_path: tuple[custom_s3_path.MyS3Path, custom_s3_path.MyS3Client]` which didn't seem to do anything in pyright for me locally), but it works!

kujenga:
great, let me know if there's anything else!

kujenga:
Hi @pjbull , any chance you'd be able to tag a release with this new functionality? Thank you!

pjbull:
@kujenga Yep! Need to get a release out for 3.13 anyway in the next few days.

PR #468, opened 2024-09-01
Implement HTTP(s) support

### LINKED ISSUE #455: Read from http - httppathlib?
### Best approach to support reading data from http via a `pathlib`-like class, i.e. `httppathlib`?

In the [pangeo](https://pangeo.io/) / [xarray](https://xarray.dev/) community we do a lot of reading of remote scientific data (particularly [netCDF](https://www.unidata.ucar.edu/software/netcdf/) and [Zarr]()). We generally want to treat 3 cases the same way: local filesystems, cloud storage, and http urls. The latter is important partly because a lot of archival scientific data is still only available from servers via http (e.g. via [openDAP urls](https://www.earthdata.nasa.gov/engage/open-data-services-and-software/api/opendap)), and we often want to pull it out and deposit it onto cloud storage (e.g. using [pangeo-forge](https://github.com/pangeo-forge/pangeo-forge-recipes)).

We currently use [`fsspec`](https://github.com/fsspec/filesystem_spec) to abstract over these different filesystems, but despite much engagement upstream we have unfortunately experienced [chronic reliability issues](https://github.com/fsspec/filesystem_spec/issues/579#issuecomment-1140377495) stemming from [ill-defined interfaces](https://github.com/fsspec/filesystem_spec/issues/1446).

CloudPathlib looks really nice, especially the strict typing and clear interface. (I'm in awe of the [`AnyPath` virtual superclass trick](https://cloudpathlib.drivendata.org/stable/api-reference/anypath/) too - and with https://github.com/drivendataorg/cloudpathlib/issues/347 would be even cooler!) The `Path` abstraction also just seems more like the minimally-useful one, rather than trying to emulate a whole filesystem. 

Rather than trying to support every filesystem under the sun as `fsspec` does, **I'm wondering if we could just use `pathlib`, `cloudpathlib`, and some new `httppathlib`?**

Do you have any thoughts on:
1. Whether you think this is a good idea?
2. The experiments already performed with xarray in https://github.com/drivendataorg/cloudpathlib/issues/96#issuecomment-1399391180?
3. How hard it might be to get a `httppathlib` to conform to the `pathlib` interface?
4. Where such a project might live: in `cloudpathlib` or in a separate repository?

pjbull:
Initial pass at http.

 - implemented with standard library
 - tested against Python `http.server`
 - will try writing/deleting files with PUT/DELETE in case your server allows it
 - will try parsing "directories" from the returned HTML; seems like this grabs the right things from python's `http.server` and also looks like it should work for apache and nginx file servers from my googling.  (allows user to override this method for their own server)
 - Bonus fix, use `.anchor` instead of `.cloud_prefix` where appropriate. For cloud providers, `.anchor` is the same as `.cloud_prefix` (e.g., `s3://`).  For http, the anchor should be the server `http://example.com` and the `.cloud_prefix` should be `http://`. Constructing new paths should use `.anchor`

Limitations/caveats:
 - ~Assumes that url must have suffix (e.g., `http://example.com/file.txt`) to be a file; if no suffix, assumes dir. This is definitely not true of real-world URLs, but maybe is an ok assumption for anything serving files?~
 - Assumes urls must end in `/` to be directories (user can pass custom test for urls if they have a different scenario)

Left to do:
 - [x] Add `https`
 - [x] Expose all the parsed url components publicly (not behind HttpPath._url).
 - [x] Update default file + dir function to be based on a trailing slash
 - [x] Allow also overriding the file vs. dir function
 - [x] Add tests for http specific functions
 - [x] Turn off noisy logs for test http server by default
 - [x] Add tests that url search strings and fragments are persisted correctly
 - [x] Documentation (table, caveats, docstrings, headers, auth (add cookies), files v dirs, etc.)
 - [x] HTTP Test suite often passes locally, but not always; need to debug and make less flaky

Closes #455


jayqi:
> Bonus fix, use `.anchor` instead of `.cloud_prefix` where appropriate. For cloud providers, `.anchor` is the same as `.cloud_prefix` (e.g., `s3://`). For http, the anchor should be the server `http://example.com` and the `.cloud_prefix` should be `http://`. Constructing new paths should use `.anchor`

We shouldn't call this "anchor". Anchor has a different meaning in a close enough context that I think this will be confusing. In particular, "anchor" colloquially refers to "anchor tags" in HTML documents, which you can reference within a URL with fragment identifiers `#someref`. The part preceding the `://` is called "scheme" in URL/URI vocabulary and we should probably stick with that if we're going to call it anything. ([Reference](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax))

> Assumes that url must have suffix (e.g., `http://example.com/file.txt`) to be a file; if no suffix, assumes dir. This is definitely not true of real-world URLs, but maybe is an ok assumption for anything serving files?

I'm not sure I like this. There are often genuine files that don't have file extensions that are usually plain text files, e.g., `README`, `LICENSE`, `Makefile`. 

My impression is that, (while it's not absolute) the most common convention and default for many web servers and frameworks is to use a trailing slash explicitly to serve directories.

pjbull:
> We shouldn't call this "anchor". Anchor has a different meaning in a close enough context that I think this will be confusing.

Hm, yeah, I can see this is potentially confusing, but we're not calling it "anchor" arbitrarily. We're looking for the right analogy to populate the existing `.anchor` pathlib property. In the [`pathlib` context `.anchor` is a `Path` object that is drive + root](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.anchor). We're using it to refer to scheme + netloc (not just the scheme) which reasoning by analogy feels about right. The difference from the cloud providers is that the scheme is the root is the drive (e.g., listing `s3://` lists buckets you have access to). IMO docs can cover this source of potential confusion without too much worry, especially since anything referring to an anchor in a url is most often called the "fragment".

> use a trailing slash explicitly to serve directories.

Yeah, this was the other default rule I considered—I think you're right it's probably a more reliable default. That is how the python server works (except for the root, which does not redirect to the slash). Also planning for the method to be overridable so people can configure for their particular servers if it is different.

MattOates:
So one thing to be aware of is your code assumes no one would ever do HttpPath("http://username:password@host/path/file.txt") for basic auth. url.netloc looks like the string "username:password@host" in this instance. Im currently struggling to see how in the Cloudpathlib design I can actually inject the auth from the URL into the client too. Looks more like you would always have to explicitly provide a client which feels not incredible from the user side.

pjbull:
> So one thing to be aware of is your code assumes no one would ever do HttpPath("http://username:password@host/path/file.txt") for basic auth.

@MattOates I guess I was thinking that this just sticks around on that path and any paths derived from it (since we don't mess with netloc) and should "just work." Does it not? Maybe there are just some bugs to clean up (sorry, haven't had a chance to poke at it and confirm/deny).

> Looks more like you would always have to explicitly provide a client which feels not incredible from the user side.

I think that we could (1) support env vars for basic auth like most of the cloud providers have for auth, (2) point people towards [how to set the default client](https://cloudpathlib.drivendata.org/stable/authentication/#advanced-use), or (3) recommend an explicit client.

jayqi:
> I guess I was thinking that this just sticks around on that path and any paths derived from it

Does that mean we're going to print out the username and password in plaintext in string representations?

pjbull:
> Does that mean we're going to print out the username and password in plaintext in string representations?

Yeah, FWIW urllib also is happy to just print that out too. We could be more conservative, but I'm not inclined to add a special case if the standard lib doesn't.

```python
In [1]: from urllib.parse import urlparse

In [2]: urlparse("http://user@pass:www.example.com")
Out[2]: ParseResult(scheme='http', netloc='user@pass:www.example.com', path='', params='', query='', fragment='')
```

pjbull:
Ok, I think this is basically all implemented and ready for review. There are two test things that I am trying to track down:

 - [ ] intermittent file not found error in CI (does not repro locally) that is probably a timing problem/race condition
 - [ ] windows tests running so slowly that the whole CI run is timing out (getting a windows machine to see if this repros or is just a ci thing)

pjbull:
OK @jayqi, this is fully working and ready for review!

tests/test_cloudpath_file_io.py
```diff
@@ -83,19 +91,19 @@ def glob_test_dirs(rig, tmp_path):
...
-        (root / "dirB" / "fileB").write_text("fileB")
+        (root / "dirB" / "fileB.txt").write_text("fileB")
         (root / "dirC").mkdir()
         (root / "dirC" / "dirD").mkdir()
-        (root / "dirC" / "dirD" / "fileD").write_text("fileD")
-        (root / "dirC" / "fileC").write_text("fileC")
-        (root / "fileA").write_text("fileA")
+        (root / "dirC" / "dirD" / "fileD.txt").write_text("fileD")
+        (root / "dirC" / "fileC.txt").write_text("fileC")
+        (root / "fileA.txt").write_text("fileA")
```
jayqi, line 99: Are we still covering cases elsewhere where a file has no extension?

tests/test_cloudpath_file_io.py
```diff
@@ -83,19 +91,19 @@ def glob_test_dirs(rig, tmp_path):
...
 
-    cloud_root = rig.create_cloud_path("glob-tests")
+    cloud_root = rig.create_cloud_path("glob-tests/")
     cloud_root.mkdir()
 
     _make_glob_directory(cloud_root)
 
-    local_root = tmp_path / "glob-tests"
+    local_root = tmp_path / "glob-tests/"
```
jayqi, line 106: I'm generally nervous about all of these tests that add trailing slashes to all of our paths. Are we at risk of changing what our tests are covering for the cloud providers? Maybe it's fine, Codecov doesn't say our coverage is going down.

cloudpathlib/http/httpclient.py
```diff
@@ -0,0 +1,201 @@
...
+            if isinstance(e, urllib.error.URLError) or e.code == 404:
+                return False
+            raise
+
+    def _move_file(self, src: HttpPath, dst: HttpPath, remove_src: bool = True) -> HttpPath:
+        # .fspath will download the file so the local version can be uploaded
+        self._upload_file(src.fspath, dst)
+        if remove_src:
+            self._remove(src)
+        return dst
```
jayqi, line 105: Makes me nervous that this is not done as a transaction. Is there a way to check PUT or DELETE support up front?

tests/test_cloudpath_file_io.py
pjbull: Will make sure we have some

tests/test_cloudpath_file_io.py
pjbull: I'll make sure we have explicit tests for that.

cloudpathlib/http/httpclient.py
jayqi: That seems worth it to me. Having a partial move feels like a surprising thing that could be confusing and could cause inconsistency or other correctness issues in certain cases.

pjbull:
Code review comments addressed and everything passing. Merging in HTTP support! 💖

PR #478, opened 2024-10-12
Make HNS check and client instantiation more reliable

### LINKED ISSUE #470: Detection of Azure hierarchical namespaces fails with `ResourceNotFoundError`
In the latest `0.19.0` version, all Azure operations fail with our Azure Blob Storage accounts with the `ResourceNotFoundError`.

The traceback leads to the `get_account_information` method used for HNS detection:
<details>
  <summary>Traceback</summary>

  ```
  File ".../PyCharm Professional Edition.app/Contents/plugins/python/helpers-pro/pydevd_asyncio/pydevd_asyncio_utils.py", line 117, in _exec_async_code
    result = func()
             ^^^^^^
  File "<input>", line 1, in <module>
  File ".../lib/python3.11/site-packages/azure/core/tracing/decorator.py", line 94, in wrapper_use_tracer
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/azure/storage/blob/_blob_service_client.py", line 245, in get_account_information
    process_storage_error(error)
  File ".../lib/python3.11/site-packages/azure/storage/blob/_shared/response_handlers.py", line 186, in process_storage_error
    exec("raise error from None")   # pylint: disable=exec-used # nosec
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
  File ".../lib/python3.11/site-packages/azure/storage/blob/_blob_service_client.py", line 243, in get_account_information
    return self._client.service.get_account_info(cls=return_response_headers, **kwargs) # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/azure/core/tracing/decorator.py", line 94, in wrapper_use_tracer
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/azure/storage/blob/_generated/operations/_service_operations.py", line 849, in get_account_info
    map_error(status_code=response.status_code, response=response, error_map=error_map)
  File ".../lib/python3.11/site-packages/azure/core/exceptions.py", line 161, in map_error
    raise error
azure.core.exceptions.ResourceNotFoundError: The specified resource does not exist.
RequestId:...
Time:2024-09-05T08:08:36.4501790Z
ErrorCode:ResourceNotFound
Content: <?xml version="1.0" encoding="utf-8"?><Error><Code>ResourceNotFound</Code><Message>The specified resource does not exist.
RequestId:...
Time:2024-09-05T08:08:36.4501790Z</Message></Error>
```
</details>

### PR DESCRIPTION
**Pull request by pjbull**
Issues #470 and #476 report problems with credentials being used that cannot determine if HNS is enabled on the storage account.

This fix does two things:
 - First, we only create a new credential if it is of the type in `_shared` that is implemented in both libraries (`SharedKeyCredentialPolicy`). Other kinds of credentials are properly isolated to a shared underlying library in the Azure SDK codebase.
 - Second, if the call to `get_account_information` results in `ResourceNotFound`, which indicates the credential does not have permission to get this metadata, we fall back to trying to see if the root dir in the container exists and has `hdi_folder=True` which should only be the case when HNS is enabled.

Closes #470

### CODE REVIEW

File: HISTORY.md [hunk 1]
```diff
@@ -1,8 +1,9 @@
 # cloudpathlib Changelog
 
-## UNRELEASED
+## v0.20.0 (2024-10-18)
 
 - Added support for custom schemes in CloudPath and Client subclases. (Issue [#466](https://github.com/drivendataorg/cloudpathlib/issues/466), PR [#467](https://github.com/drivendataorg/cloudpathlib/pull/467))
+- Add check for HNS that works in more scenarios and with different kinds of credentials. (Issue [#470](https://github.com/drivendataorg/cloudpathlib/issues/470), Issue [#476](https://github.com/drivendataorg/cloudpathlib/issues/476), PR [#478](https://github.com/drivendataorg/cloudpathlib/pull/478))
```
jayqi, line 6: I think this is the wrong approach to writing changelog entries. I think it should be something like: 

> Fixed errors that could occur when instantiating `AzureBlobClient` under certain authentication approaches. 

What you have right now describes the implementation detail. The thing that matters for users is the runtime errors. I'm imagining a user who is using v0.19.0 and gets an error, and can see in the changelog for v0.20.0 that it might be fixed. (It would be better if we identify what those errors are, but maybe this would be annoying to actually list out.)

jayqi (approved):
@pjbull looks good except for how this is documented in the changelog. I don't need to review again once you address that.

PR #480, opened 2024-10-22
Fix CloudPath.__truediv__ to return NotImplemented

jayqi:
Return `NotImplemented` for `CloudPath.__truediv__` in order to support operands that have implemented an `__rtruediv__` that works with `CloudPath`. 

Addresses an issue regarding compatibility with pandas-path surfaced in this issue:

- https://github.com/drivendataorg/pandas-path/issues/12#issuecomment-2427816667

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,9 @@
 # cloudpathlib Changelog
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


pjbull (approved):
Looks good, thanks!

PR #487, opened 2024-11-21
fix(az): catch HttpResponseError in _check_hns

### LINKED ISSUE #486: Azure: `_check_hns` throws an `HttpResponseError`
It seems that the `self.service_client.get_account_information()` can throw an `HttpResponseError` when the client is supplied with a `DefaultAzureCredential`:
```
azure.core.exceptions.HttpResponseError: This request is not authorized to perform this operation using this permission.
RequestId:XXXXXXXXXXXXXXXXXX
Time:2024-11-21T13:35:08.6097052Z
ErrorCode:AuthorizationPermissionMismatch
Content: <?xml version="1.0" encoding="utf-8"?><Error><Code>AuthorizationPermissionMismatch</Code><Message>This request is not authorized to perform this operation using this permission.
RequestId:8XXXXXXXXXXXXXXXXXX
Time:2024-11-21T13:35:08.6097052Z</Message></Error>
```

Maybe the `except ResourceNotFoundError` should be extended to catch the `HttpResponseError` as well?

M0dEx:
Fixed an uncaught exception in `check_hns`, which can fail with the `HttpResponseError` when supplied with a `DefaultAzureCredential`.

Closes #486.

### DIFF
--- a/cloudpathlib/azure/azblobclient.py
+++ b/cloudpathlib/azure/azblobclient.py
@@ -1,9 +1,11 @@
 from datetime import datetime, timedelta
 import mimetypes
 import os
+from http import HTTPStatus
 from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union
 
+from typing_extensions import cast
 
 from ..client import Client, register_client_class
 from ..cloudpath import implementation_registry
@@ -13,7 +15,7 @@ from .azblobpath import AzureBlobPath
 
 
 try:
-    from azure.core.exceptions import ResourceNotFoundError
+    from azure.core.exceptions import HttpResponseError, ResourceNotFoundError
     from azure.core.credentials import AzureNamedKeyCredential
 
     from azure.storage.blob import (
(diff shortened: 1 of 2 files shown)


M0dEx:
Also, while going through the code, it occurred to me that `_check_hns` should probably be cached on a account/container level, as it can slow down a lot of operations.

cloudpathlib/azure/azblobclient.py
```diff
@@ -202,7 +202,7 @@ def _check_hns(self, cloud_path: AzureBlobPath) -> Optional[bool]:
             try:
                 account_info = self.service_client.get_account_information()  # type: ignore
                 self._hns_enabled = account_info.get("is_hns_enabled", False)  # type: ignore
-            except ResourceNotFoundError:
+            except (ResourceNotFoundError, HttpResponseError):
```
pjbull, line 205: Per the stack trace in #486, can we confirm the error code is `ErrorCode:AuthorizationPermissionMismatch` and if not, re-raise the `HttpResponseError`? That will make other potential errors easier to track down.

pjbull (changes requested):

cloudpathlib/azure/azblobclient.py
M0dEx: Is checking for `HTTPStatus.FORBIDDEN` sufficient?

PR #495, opened 2024-12-27
Azure support for `..` in path

### LINKED ISSUE #494: Azure client fail when path contains `..`
When `..` are in the path, the AzureBlobPath fails to list directory with

```
azure.core.exceptions.ResourceNotFoundError: The specified path does not exist.
```

Interestingly, the metadata to identify whether a path exists, is dir, etc., works just fine.

lejmr:
When looking into #494, I found several possible solutions.

1) The proposed one (resolve correctly blob path)
2) Create alternative property, say `AzureBlobPath.blob_resolved` and use in on line https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/azure/azblobclient.py#L337

I tested the code is working in my specific situation and all tests are still passing.

Closes #494


pjbull:
Closing this PR per discussion in #494

PR #516, opened 2025-04-13
Rerun flaky tests when live cloud

jayqi:
Use [pytest-rerunfailures](https://github.com/pytest-dev/pytest-rerunfailures) to rerun flaky tests when running on live cloud services.


pjbull (approved):
Thanks!

PR #547, opened 2026-02-04
Fix S3 copy/move extra args filtering for SSE-C

### LINKED ISSUE #500: S3 move and copy operations with SSE-C
We use S3 and SSE-C to encrypt the objects stored in the remote S3 repository.
I've initialized the `S3Client` accordingly with:
```
extra_args={
    "SSECustomerKey": sse_c_key,
    "SSECustomerAlgorithm": "AES256",
    "CopySourceSSECustomerKey": sse_c_key,
    "CopySourceSSECustomerAlgorithm": "AES256",
}
```

Reading and writing objects worked that way but copying and moving (= copy and delete) do not: the `CopySourceSSECustomerKey` and `CopySourceSSECustomerAlgorithm` are not passed to the copy operation.

If I saw it correctly, the reason is that in [s3client.py:298](https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/s3/s3client.py#L298) the property `boto3_ul_extra_args` is used, which ensures that all extra arguments match the list `S3Transfer.ALLOWED_UPLOAD_ARGS` (coming from `TransferManager.ALLOWED_DOWNLOAD_ARGS`). Unfortunately, the `Copy*` attributes are not included there but in `TransferManager.ALLOWED_COPY_ARGS`.

Would it be possible to use the `ALLOWED_COPY_ARGS` there? Or is this restricted intentionally?

jiangmuran:
## Summary
- filter copy ExtraArgs with boto3 copy-allowed keys so SSE-C CopySource parameters are forwarded
- use copy-safe extra args for both copy and copy_from paths in S3 moves

## Related Issues
- fixes #500

## Testing
- not run (requires S3 setup)

### DIFF
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -11,7 +11,7 @@ from .s3path import S3Path
 
 try:
     from boto3.session import Session
-    from boto3.s3.transfer import TransferConfig, S3Transfer
+    from boto3.s3.transfer import TransferConfig, S3Transfer, TransferManager
     from botocore.config import Config
     from botocore.exceptions import ClientError
     import botocore.session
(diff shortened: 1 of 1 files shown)


pjbull:
superseded https://github.com/drivendataorg/cloudpathlib/pull/576

PR #549, opened 2026-02-25
s3: pass copy extra args during move and copy operations

ai-m-kirk-51517:
This PR fixes issue #500 where `Copy*` attributes (like `CopySourceSSECustomerKey`) were not being passed to S3 copy operations during move/copy. 

I have introduced `boto3_copy_extra_args` which uses `S3Transfer.ALLOWED_COPY_ARGS` to filter relevant extra arguments and ensured they are passed to the `target.copy` call.

### Checklist
- [x] Code change follows the logic identified in issue #500.
- [x] Uses correct boto3/S3Transfer constants.
- [x] Verified that `_move_file` now uses the specific copy extra args.


pjbull:
superseded https://github.com/drivendataorg/cloudpathlib/pull/576

PR #555, opened 2026-04-01
Fix SSE-C copy operations: pass CopySourceSSECustomerKey to S3 copy

### LINKED ISSUE #500: S3 move and copy operations with SSE-C

xingzihai:
Issue: #500 - Failing S3 move and copy operations with SSE-C

When using S3 with SSE-C encryption, copy and move operations fail because CopySourceSSECustomerKey parameters are not passed.

Changes:
- Add boto3_copy_extra_args for copy-specific SSE-C params
- Update _move_file to use upload + copy-specific args

Fixes #500

### DIFF
--- a/cloudpathlib/s3/s3client.py
+++ b/cloudpathlib/s3/s3client.py
@@ -119,6 +119,21 @@ class S3Client(Client):
             k: v for k, v in extra_args.items() if k in S3Transfer.ALLOWED_UPLOAD_ARGS
         }
 
+        # copy ops need SSE-C args for both source (CopySource*) and destination (SSE*)
+        # https://docs.aws.amazon.com/AmazonS3/latest/userguide/copysource-encrypted-sse-c.html
+        self.boto3_copy_extra_args = {
+            k: v
+            for k, v in extra_args.items()
+            if k in [
+                "CopySourceSSECustomerAlgorithm",
+                "CopySourceSSECustomerKey",
+                "CopySourceSSECustomerKeyMD5",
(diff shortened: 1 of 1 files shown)


pjbull:
superseded by https://github.com/drivendataorg/cloudpathlib/pull/576

Issue #439, opened 2024-06-04
Unable to load files with a colon in filename (Windows Client only)


glusin:
Filenames on cloud are allowed to contain colons, but it is not allowed in Windows filenames.
When a file is downloaded from a cloud it is first saved in a temp file. If filename contains colons then the code fails

pjbull:
Thanks for the issue @glusin.

What were you trying to do when you encountered the issue? Can you provide a short, reproducible example and the stack trace that accompanies the error?

glusin:
We have a class to handle blob:// paths that inherits from `AzureCloudPath`.
For az:// the example would look like this:

```
from cloudpathlib import CloudPath, AzureBlobClient

client = AzureBlobClient(
    account_url="account_url",
    credential=credentials,
)

file_path = CloudPath('az://path/to/a/file:containing:colons', client=client)
file_path.open()  # crashes here
```


This is a traceback:
```

    File "C:\Users\PawelGlusinski\PycharmProjects\project\temp.py", line 5, in <module>
    file_path.open()
  File "C:\Users\PawelGlusinski\AppData\Local\pypoetry\Cache\virtualenvs\project-_1bmJiI5-py3.11\Lib\site-packages\cloudpathlib\cloudpath.py", line 559, in open
    self._refresh_cache(force_overwrite_from_cloud=force_overwrite_from_cloud)
  File "C:\Users\PawelGlusinski\AppData\Local\pypoetry\Cache\virtualenvs\project-_1bmJiI5-py3.11\Lib\site-packages\cloudpathlib\cloudpath.py", line 1121, in _refresh_cache
    self.download_to(self._local)
  File "C:\Users\PawelGlusinski\AppData\Local\pypoetry\Cache\virtualenvs\project-_1bmJiI5-py3.11\Lib\site-packages\cloudpathlib\cloudpath.py", line 900, in download_to
    return self.client._download_file(self, destination)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\PawelGlusinski\AppData\Local\pypoetry\Cache\virtualenvs\project-_1bmJiI5-py3.11\Lib\site-packages\cloudpathlib\azure\azblobclient.py", line 140, in _download_file
    with partial_local_path.open("wb") as data:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #557, opened 2026-04-09
Add support for S3 Multi-Region Access Point (MRAP) URLs

### LINKED ISSUE #556: Add support for S3 Multi-Region Access Point (MRAP) URLs
## Description

S3 [Multi-Region Access Points (MRAPs)](https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiRegionAccessPoints.html) use ARN-based URLs instead of a plain bucket name:

```
s3://arn:aws:s3::123456789012:accesspoint/my-mrap.mrap/path/to/file.txt
```

Currently, `S3Path` parses the `bucket` by splitting on the first `/` after `s3://`, which gives `arn:aws:s3::123456789012:accesspoint` — cutting off the `.mrap` suffix. This causes `bucket` and `key` to be computed incorrectly, breaking all boto3 calls since they receive the wrong `Bucket` and `Key` parameters.

## Current behavior

```python
p = S3Path("s3://arn:aws:s3::123456789012:accesspoint/my-mrap.mrap/folder/file.txt")
p.bucket  # "arn:aws:s3::123456789012:accesspoint"  ← wrong
p.key     # "my-mrap.mrap/folder/file.txt"          ← wrong
```

## Expected behavior

```python
p.bucket  # "arn:aws:s3::123456789012:accesspoint/my-mrap.mrap"  ← full ARN
p.key     # "folder/file.txt"                                     ← correct
```

## Proposed fix

Override the `bucket` property on `S3Path` to detect MRAP ARNs via a regex and return the full ARN when matched, falling back to the existing behaviour for regular bucket names.

fedemengo:
Closes #556

Changes

  - `cloudpathlib/s3/s3path.py`: added _MRAP_PATTERN regex and overridden bucket property on S3Path
  - `tests/mock_clients/mock_s3.py`: removed bucket name check from head_object (the mock never routes by bucket, only by key)
  - `tests/test_s3_specific.py`: added 3 tests covering bucket/key parsing, path manipulation, and end-to-end file operations via the
  mock backend

  Note: using MRAP paths requires `botocore[crt]` for SigV4a signing:
  
`pip install botocore[crt]`

----------------

Contributor checklist:

 - [ ] Any generic new functionality is replicated across cloud providers if necessary
 - [x] Linting passes locally
 - [x] Tests pass locally

### DIFF
--- a/cloudpathlib/s3/s3path.py
+++ b/cloudpathlib/s3/s3path.py
@@ -83,9 +83,13 @@ class S3Path(CloudPath):
 
         :type: :class:`str`
         """
+        if hasattr(self, "_bucket"):
+            return self._bucket
         if match := _MRAP_PATTERN.match(str(self)):
-            return match.group("arn")
-        return self._no_prefix.split("/", 1)[0]
+            self._bucket = match.group("arn")
+        else:
+            self._bucket = self._no_prefix.split("/", 1)[0]
+        return self._bucket
 
     @property
     def key(self) -> str:
--- a/cloudpathlib/s3/s3path.py
+++ b/cloudpathlib/s3/s3path.py
@@ -1,5 +1,6 @@
 import os
 import re
+import sys
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any, Optional, TYPE_CHECKING
@@ -32,6 +33,7 @@ class S3Path(CloudPath):
     cloud_prefix: str = "s3://"
     client: "S3Client"
     _bucket: str
+    _local_path: Path
 
     @property
     def drive(self) -> str:
@@ -106,3 +108,14 @@ class S3Path(CloudPath):
     @property
     def etag(self):
         return self.client._get_metadata(self).get("etag")
+
+    @property
+    def _local(self) -> Path:
+        if hasattr(self, "_local_path"):
+            return self._local_path
+        no_prefix = self._no_prefix
+        # `:` is invalid in Windows paths; percent-encode it for MRAP ARNs
+        if sys.platform == "win32":
+            no_prefix = no_prefix.replace(":", "%3A")
+        self._local_path = self.client._local_cache_dir / no_prefix
+        return self._local_path
--- a/cloudpathlib/s3/s3path.py
+++ b/cloudpathlib/s3/s3path.py
@@ -1,4 +1,5 @@
 import os
+import re
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any, Optional, TYPE_CHECKING
@@ -7,6 +8,10 @@ from ..cloudpath import CloudPath, NoStatError, register_path_class
 
 if TYPE_CHECKING:
     from .s3client import S3Client
+
+_MRAP_PATTERN = re.compile(
+    r"^s3://(?P<arn>arn:aws:s3::\d{12}:accesspoint/[^/]+\.mrap)(?:/(?P<key>.*))?$"
+)
 
 
 @register_path_class("s3")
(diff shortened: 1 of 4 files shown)


cloudpathlib/s3/s3path.py
```diff
@@ -74,6 +79,12 @@ def stat(self, follow_symlinks=True):
 
     @property
     def bucket(self) -> str:
+        """The bucket name, or the full MRAP ARN for MRAP paths.
+
+        :type: :class:`str`
+        """
+        if match := _MRAP_PATTERN.match(str(self)):
+            return match.group("arn")
```
pjbull: Let's cache this since it shouldn't change.

fedemengo:
the changes were simple enough, @pjbull gentle ping for review

tests/mock_clients/mock_s3.py
```diff
@@ -220,14 +220,9 @@ def list_buckets(self):
         return {"Buckets": [{"Name": DEFAULT_S3_BUCKET_NAME}]}
 
     def head_object(self, Bucket, Key, **kwargs):
-        if (
-            not (self.root / Key).exists()
-            or (self.root / Key).is_dir()
-            or Bucket != DEFAULT_S3_BUCKET_NAME
```
pjbull, line 226: IIRC some of the tests use this check in some scenarios. Can we also just add a simple check that it doesn't have '.mrap' here

pjbull (changes requested):
A couple comments.Also is it possible to update the tests to use s3_rig following the pattern of the existing tests rather than using `S3Path` directly (as in some tests) or using `monkeypatch` as in the others.

fedemengo:
I replaced S3Path with rig_ops when tests is not purly path parsing

fedemengo:
@pjbull I replaced S3Path with rig_ops when tests are not purly path parsing

pjbull:
@fedemengo see linting errors in the action logs.

fedemengo:
> @fedemengo see linting errors in the action logs.

@pjbull I had missed that https://github.com/drivendataorg/cloudpathlib/pull/557/changes/8daee30fab1029ea9f1c4ff2a91e06326e3a807d

pjbull:
@fedemengo tests are failing on windows because `:` is not a valid character in the filepath.

We're coming up against this separate issue:
https://github.com/drivendataorg/cloudpathlib/issues/439

Maybe for now you can override the [`_local`](https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/cloudpath.py#L1450-L1453) property on `S3Path` and if there is a `:` and we are on Windows, we percent-encode the `:`?

fedemengo:
@pjbull something like this https://github.com/drivendataorg/cloudpathlib/pull/557/changes/baf833ec3ccbc54bfaa31f97dcde330e2603647f

pjbull:
Test needs an update to strip `.drive` from the path before checking no `:`:.

```
FAILED tests/test_s3_specific.py::test_mrap_local_path_windows_encoding - AssertionError: Colon found in local path on simulated Windows: C:\Users\RUNNER~1\AppData\Local\Temp\tmpc_bmszu4\arn%3Aaws%3As3%3A%3A123456789012%3Aaccesspoint\my-mrap.mrap\some\key.txt
assert ':' not in 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpc_bmszu4\\arn%3Aaws%3As3%3A%3A123456789012%3Aaccesspoint\\my-mrap.mrap\\some\\key.txt'
```

fedemengo:
@pjbull ok hopefully this is it, if there is another failure I'll probably spin up a  Windows vm

pjbull:
@fedemengo linting failure

fedemengo:
@pjbull my bad, ran the linter ac20cb4ef06f041e85404c5982708af9cf133376

PR #560, opened 2026-04-29
Add support for S3 Multi-Region Access Point (MRAP) URLs (#557)

pjbull:
Live tests for #557

* Add MRAP URL support

* test MRAP

* update history

* use walrus operator

* cache bucket parsing

* address pr review

* fix linter issues

* url escape `:` on win and cache

* handle Windows drive in test assertion

* make lint


fedemengo:
MRAP does require `botocore[crt]`

from #557

> Note: using MRAP paths requires botocore[crt] for SigV4a signing:
> 
> pip install botocore[crt]

The failure looks consistent with that missing dependency. Not sure if we should treat this as optional dep or not

PR #565, opened 2026-06-09
Use HeadObject instead of GetObject for S3 metadata

### LINKED ISSUE #564: `S3Client._get_metadata` uses GetObject where HeadObject would do
`.stat()`, `.etag`, and the metadata they feed all go through `S3Client._get_metadata`, which reads object metadata via a **GetObject** instead of a **HeadObject**. In `cloudpathlib/s3/s3client.py`:

https://github.com/drivendataorg/cloudpathlib/blob/1d65750d99ac320b8f46016322d6b8fe1995d90f/cloudpathlib/s3/s3client.py#L137-L149

`ObjectSummary.get()` issues a GetObject. botocore returns the body as a lazy `StreamingBody`, which `_get_metadata` never reads — so the bytes aren't pulled into Python — but it's still a GetObject (the server treats it as an object read), and since the body is never consumed, the underlying HTTP connection can't be returned to the pool and gets torn down rather than reused. All five fields the method reads are present on a HeadObject response, which has no body at all.

**Why this matters generally** (my own trigger is niche): the other backends fetch metadata with a metadata-only request rather than an object read — Azure `get_blob_properties()`, GCS `get_blob()` (an `objects.get` metadata GET, not a media download), local `Path.stat()`. S3 is the odd one out. HeadObject makes `.stat()`/`.etag`/`.size` a lighter request and brings S3 in line with the rest of the library.

My niche case: I'm on an S3-compatible gateway (lakeFS behind an edge that gzips compressible bodies and returns them chunked). That drops `Content-Length` from the GET response, so `data["ContentLength"]` raises `KeyError`. HeadObject keeps `Content-Length`.

**Suggested fix** — swap the GetObject for a HeadObject. The module already uses `head_object` in `_s3_file_query`, the same `boto3_dl_extra_args` are valid there, and HeadObject needs the same `s3:GetObject` permission, so no IAM change:

```python
def _get_metadata(self, cloud_path: S3Path) -> Dict[str, Any]:
    data = self.client.head_object(
        Bucket=cloud_path.bucket,
        Key=cloud_path.key,
        **self.boto3_dl_extra_args,
    )
    return {
        "last_modified": data["LastModified"],
        "size": data["ContentLength"],
        "etag": data["ETag"],
        "content_type": data.get("ContentType", None),
        "extra": data["Metadata"],
    }
```

One behaviour note for the PR: GetObject raises `NoSuchKey` on a missing key, but HeadObject (no response body) raises a generic 404 `ClientError`, so `S3Path.stat()`'s `except NoSuchKey` would need to also handle the 404 (and the test mock's `head_object` needs the same field set). This isn't new ground — `_s3_file_query` already catches `(ClientError, NoSuchKey)` for its own `head_object` call. Happy to send a PR with those bundled if it's welcome.

mattijsdp:
Closes #564

S3Client._get_metadata read object metadata via ObjectSummary.get() (a GetObject), which issues a full object read and leaves an unread response stream instead of returning the connection to the pool. Switch to head_object, which reads only the metadata and keeps Content-Length on S3-compatible gateways that drop it from GetObject responses, while bringing S3 in line with the other backends.

Since head_object returns no body, a missing key surfaces as a generic 404 ClientError rather than NoSuchKey, so S3Path.stat now translates a 404 to NoStatError and lets other errors (e.g. 403) propagate. Updates the S3 test mock and adds regression tests.

----------------

Contributor checklist:

 - [x] Linting passes locally
 - [x] Tests pass locally

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 ## UNRELEASED
 
 - Fixed mypy 2.x type errors in `Client` and `CloudPath` that caused CI lint failures (Issue [#563](https://github.com/drivendataorg/cloudpathlib/issues/563), PR [#566](https://github.com/drivendataorg/cloudpathlib/pull/566))
+- Changed `S3Client._get_metadata` to read object metadata with `HeadObject` instead of `GetObject`, so `stat`, `etag`, and `size` no longer open the object body. Also fixes a `KeyError` on `ContentLength` against S3-compatible gateways that drop `Content-Length` from `GetObject` responses. (Issue [#564](https://github.com/drivendataorg/cloudpathlib/issues/564), PR [#TODO])
 
 ## v0.24.0 (2026-04-29)
 - Added support for S3 Multi-Region Access Point (MRAP) URLs in `S3Path` (Issue [#556](https://github.com/drivendataorg/cloudpathlib/issues/556), PR [#557](https://github.com/drivendataorg/cloudpathlib/pull/557))
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,7 +3,7 @@
 ## UNRELEASED
 
 - Fixed mypy 2.x type errors in `Client` and `CloudPath` that caused CI lint failures (Issue [#563](https://github.com/drivendataorg/cloudpathlib/issues/563), PR [#566](https://github.com/drivendataorg/cloudpathlib/pull/566))
-- Changed `S3Client._get_metadata` to read object metadata with `HeadObject` instead of `GetObject`, so `stat`, `etag`, and `size` no longer open the object body. Also fixes a `KeyError` on `ContentLength` against S3-compatible gateways that drop `Content-Length` from `GetObject` responses. (Issue [#564](https://github.com/drivendataorg/cloudpathlib/issues/564), PR [#TODO])
+- Changed `S3Client._get_metadata` to read object metadata with `HeadObject` instead of `GetObject`, so `stat`, `etag`, and `size` no longer open the object body. Also fixes a `KeyError` on `ContentLength` against S3-compatible gateways that drop `Content-Length` from `GetObject` responses. (Issue [#564](https://github.com/drivendataorg/cloudpathlib/issues/564), PR [#565](https://github.com/drivendataorg/cloudpathlib/pull/565))
 
 ## v0.24.0 (2026-04-29)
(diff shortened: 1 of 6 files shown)


cloudpathlib/s3/s3client.py
```diff
@@ -135,9 +135,11 @@ def __init__(
...
 
     def _get_metadata(self, cloud_path: S3Path) -> Dict[str, Any]:
-        # get accepts all download extra args
-        data = self.s3.ObjectSummary(cloud_path.bucket, cloud_path.key).get(
-            **self.boto3_dl_extra_args
+        # head_object accepts all download extra args and reads metadata without the body
+        data = self.client.head_object(
+            Bucket=cloud_path.bucket,
+            Key=cloud_path.key,
+            **self.boto3_dl_extra_args,
```
pjbull, line 142: Does `head_object` respect what we collect in these kwargs?

HISTORY.md
```diff
@@ -1,5 +1,9 @@
 # cloudpathlib Changelog
 
+## UNRELEASED
+
+- Changed `S3Client._get_metadata` to read object metadata with `HeadObject` instead of `GetObject`, so `stat`, `etag`, and `size` no longer open the object body. Also fixes a `KeyError` on `ContentLength` against S3-compatible gateways that drop `Content-Length` from `GetObject` responses. (Issue [#564](https://github.com/drivendataorg/cloudpathlib/issues/564), PR [#TODO])
```
pjbull: Add pr number

pjbull (commented):
Generally looking good, a few quick comments in there.

Did you test against your s3 compatible server and real s3?

Also, did you make sure linting passes?

cloudpathlib/s3/s3client.py
mattijsdp: Yes, checked all 7 (see https://docs.aws.amazon.com/boto3/latest/reference/services/s3/client/head_object.html)

HISTORY.md
mattijsdp: Done

mattijsdp:
Yes, linting passed and checked on real s3 bucket and s3 compatible server (lakefs).

mattijsdp:
Code quality errors are unrelated to my changes I think.

pjbull:
Looks like linter package version updates. I'll get it fixed on main so you can rebase.

pjbull:
@mattijsdp fixed on the main branch if you want to rebase this. Thanks for bearing with.

mattijsdp:
Rebased onto latest `main` (now includes #566). Lint (`black`/`flake8`/`mypy`) and the test suite pass locally. Thanks for the quick fix @pjbull!

pjbull:
live tests expected to fail on fork, moving to repo-local branch to test real backends.

